PR #115, opened 2021-10-07
Feature/add encrypted storage option

dsamojlenko:
# Summary | Résumé

Adds ability to store/retrieve encrypted values from the `wp_options` table. Specifically for use in storing the Notify API key and related secrets for collections/sites, rather than storing them in the environment.

## CDS\EncryptedOption
This class is uses the [Laravel Encryption](https://laravel.com/docs/8.x/encryption) library to encrypt and decrypt strings. It includes methods to get and set Encrypted Options, as well as convenience methods for encrypting and decrypting strings directly. Also included is a CLI command to generate an Encryption Key (see below).

## Encryption Key
There is a new mandatory environment variable, ENCRYPTION_KEY. This is a base64 encoded randomized string that is used to encode and decode payloads. 

There is a default key set in wp-config.php. It is fine to use this key for local and dev environments, but a unique key *must* be generated for each production/staging environment.

## Cli command
To generate an encryption key, ensure you have the docker-compose environment running, and use the npm script:

```
npm run generate-encryption-key
```

Alternatively, you can docker exec into the CLI container and run the wp-cli command directly:

```
➜  ~ docker exec -it cli zsh
(devcontainer) ➜  ~ cd /var/www/html
(devcontainer) ➜  html wp generate-encryption-key
Success: Here is an encryption key, you should add it to your .env file as ENCRYPTION_KEY
base64:umHzRQa64IkwSegQFdezRoWktEJhi1bP9OmMPTWqp8I=
```

## Notify API Settings

Also adds a settings page for Notify API related settings. There are five fields, four of which are currently treated as sensitive. The four sensitive fields are encrypted in the database and obfuscated when displayed.

| Unfilled  | Filled |
| ------------- | ------------- |
| image  | image |

## TODO

- [x] Write tests
- [x] Use the stored API keys
- [x] ~~Remove replaced Environment Variables from infra (will do in future PR)~~
- [x] Update README


timarney:
Nice work @dsamojlenko

PR #172, opened 2021-10-21
Change local docker environment to use Alpine PHP-FPM and Apache

dsamojlenko:
# Summary | Résumé

Changing the docker base image to use Alpine. An earlier iteration of this included both PHP-FPM and Apache in the same container, using supervisord to keep both processes running. We've since decided to move towards a split with an Apache container acting as reverse-proxy to the wordpress/php-fpm container.

This PR contains the changes to the docker-compose development environment for this split. A future PR will introduce an Apache "sidecar" to our ECS WordPress container.

## Change to vendor dependencies config
Note that included here is a change to how our custom Plugin and Theme are configured to manage their dependencies. 

They have been reconfigured as "path" dependencies in `wordpress/composer.json` and all of their dependencies will now be installed in `wordpress/vendor`. 

This allows Composer to resolve any conflicts between our installed and custom plugins and themes.

Note that this required downgrading a dependency of the Theme, `paquettg/php-html-parser` as the newer version of this package required a version of Guzzle that conflicted with the version required by two of our plugins.

## Testing

There are substantial changes to the underlying containers and volumes in this PR, so best bet is to do a full cleanup of your local environment. You can stop and delete the `gc-articles` app in Docker Desktop, and manually delete all the underlying volumes.

Alternatively, you can just wipe out all your local volumes and apps:

```
docker system prune -a --volumes
```

With your local environment cleaned up, ensure you have this PR checked out and your .env file configured (there are no new environment variables, so you should be fine with your old one), run the new setup steps:

```
docker-compose run composer
docker-compose run install
```

This will ensure your composer dependencies have been installed, and that your database is ready to go. Now you can bring up the rest of the environment:

```
docker-compose up
```

## Known errors at startup

When running the `docker-compose run install` command, you will note a bunch of errors output. This is known, it has to do with the timing of a bunch of our hooks that attempt to interact with the database before it's ready. It will not affect the rest of the experience, we'll fix it later.


README.md
```diff
@@ -4,10 +4,10 @@
 
 Docker-compose and VS Code Remote Container environment featuring:
 
+- apache (reverse-proxy to wordpress/php-fpm)
 - mariadb (instead of mysql)
-- wordpress (wordpress/apache)
+- wordpress (wordpress/php-fpm)
 - wp-cli & composer (devcontainer)
-- mailhog (fake mail)
```
dsamojlenko, line 10: this was removed previously

wordpress/composer.json
```diff
@@ -47,19 +57,12 @@
...
   },
   "scripts": {
-    "post-install-cmd": [
-      "cd wp-content/themes/cds-default && composer install",
-      "cd wp-content/mu-plugins/cds-base && composer install"
-    ],
-    "post-update-cmd": [
-      "cd wp-content/themes/cds-default && composer update",
-      "cd wp-content/mu-plugins/cds-base && composer update"
-    ],
```
dsamojlenko, line 62: Don't need this anymore as all dependencies will be installed in `wordpress/vendor`

wordpress/composer.json
```diff
@@ -29,7 +37,9 @@
     "wpackagist-plugin/wordpress-importer":"0.6.4",
     "wpackagist-plugin/wordpress-seo": "^16.9",
     "humanmade/s3-uploads": "^3.0",
-    "wpackagist-plugin/gutenberg":"11.9.1"
+    "wpackagist-plugin/gutenberg":"11.9.1",
+    "cds-snc/cds-base": "@dev",
+    "cds-snc/cds-default": "@dev"
```
dsamojlenko, line 42: Treat our plugin/theme like any other composer-loaded theme/plugin, this is what gets all our dependencies in one place (see also the "path" entries in the "repositories" section above)

pcraig3 (approved):
Tried this locally and it worked. Happy to get this in sooner rather than later. 🤙

PR #203, opened 2021-10-28
Rename GC Admin role to administrator for compatibility with WPML

dsamojlenko:
# Summary | Résumé

The upcoming version of WPML 4.5 has a hard-coded reference to the `administrator` user role.

This PR changes our gcadmin role to use the `administrator` machine name. This way we'll be compatible with WPML, but also simplifies our wp-env setup since wp-env relies on the `administrator` role as well.

Also created a custom capability for managing Notify settings - previously was using manage_options, but gcadmin should not be able to access other options pages.

**NOTE** Once this is deployed, the gcadmin/GC Admin role will be removed, and any users with that role will need to be re-assigned the new administrator/GC Admin role.


pcraig3 (approved):
Everything works! Great job.

PR #205, opened 2021-10-29
Remove `fullscreen` wp icon

timarney:
Closes https://github.com/cds-snc/gc-articles-issues/issues/52

Opted to "hide" / vs replace 

Before:
Screen Shot 2021-10-29 at 1 08 14 PM

After:

Screen Shot 2021-10-29 at 1 09 44 PM


pcraig3 (commented):
I can't find a way to make it back from the editing screen now (other than the browser back button). Feels like a bridge too far for me, in terms of lost usability.

Happy to be outvoted, but that's my feeling about it.

timarney:
True - @pcraig3 I'll do a replacement

timarney:
@pcraig3 - I replaced the icon with the same dashicon used for the Admin logo

Screen Shot 2021-11-01 at 9 07 44 AM

pcraig3 (commented):
I think this PR needs to be rebased first (?), it looks like it's also including changes from my last 2 PRs (#202, #204) in addition to replacing the icon. 

Works well though. Will approve once that's done.

pcraig3 (approved):
Looks great!

PR #208, opened 2021-11-01
Add login lockdown

timarney:
Closes https://github.com/cds-snc/gc-articles-issues/issues/1

Adds login lockdown plugin 

- Adding this plugin for further evaluation.  It's lightweight vs some of the other "all in one" security plugins.  The code is in a single file (not much code) so it will be fairly easy to modify as need be.
- Covers the functionality we need exposed by a settings page

### DIFF
--- a/wordpress/.snyk
+++ b/wordpress/.snyk
@@ -12,4 +12,9 @@ ignore:
         reason: None Given
         expires: 9999-12-30T00:00:00.000Z
         created: 2021-10-27T21:19:45.920Z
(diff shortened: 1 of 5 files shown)


pcraig3 (commented):
If I install this manually it works, but when I try to install it using our scripts, I don't actually see it show up in my WordPress. 

```
- cd wordpress
- composer install
- cd ..
- docker compose up --build
```

Is there an activation step or something else that I'm not using?

timarney:
Missed a file -- updated now.

pcraig3 (approved):
Works! Looks good 👍

PR #209, opened 2021-11-01
Add button styles / color palette

timarney:
Adds initial code to ensure the Admin shows colours that match what we expect in the theme.  This will come into play more later as we start to tackle https://github.com/cds-snc/gc-articles-issues/issues/18

Screen Shot 2021-11-01 at 1 22 36 PM

### DIFF
--- a/wordpress/wp-content/themes/cds-default/functions.php
+++ b/wordpress/wp-content/themes/cds-default/functions.php
@@ -87,6 +87,59 @@ if (!function_exists('cds_setup')) {
             'style',
             'script',
         ]);
+
+        add_theme_support( 'editor-color-palette', array(
+            array(
+                'name'  => __( 'Black', 'cds-snc' ),
+                'slug'  => 'black',
+                'color'	=> '#000000',
+            ),
+            array(
+                'name'  => __( 'White', 'cds-snc' ),
+                'slug'  => 'white',
+                'color'	=> '#ffffff',
+            ),
+            array(
+                'name'  => __( 'Primary Blue', 'cds-snc' ),
+                'slug'  => 'primary-blue',
+                'color'	=> '#284162',
+            ),
+            array(
+                'name'  => __( 'Dark Blue', 'cds-snc' ),
+                'slug'  => 'dark-blue',
+                'color'	=> '#32373c',
+            ),
+            array(
+                'name'  => __( 'Light Blue', 'cds-snc' ),
+                'slug'  => 'light-blue',
+                'color' => '#b2e3ff',
+            ),
+            array(
+                'name'  => __( 'Dark Grey', 'cds-snc' ),
+                'slug'  => 'dark-grey',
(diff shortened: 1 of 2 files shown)


wordpress/wp-content/themes/cds-default/style.css
```diff
@@ -6,3 +6,209 @@
  * Domain Path: /languages
  * Version: 1.8.0
 */
+
+/* https://jsonformatter.org/sass-to-css */
+
+/*
+$colors: (
```
timarney, line 13: Noting: I didn't want to add a SASS build step in this PR so the commented out code is what is used to generate the  CSS at the bottom of the file.  Using this tool for now https://jsonformatter.org/sass-to-css

wordpress/wp-content/themes/cds-default/functions.php
```diff
@@ -87,6 +91,59 @@ function cds_setup(): void
...
+            array(
+                'name'  => __('White', 'cds-snc'),
+                'slug'  => 'white',
+                'color' => '#ffffff',
+            ),
+            array(
+                'name'  => __('Primary Blue', 'cds-snc'),
+                'slug'  => 'primary-blue',
+                'color' => '#284162',
+            ),
```
pcraig3, line 110: Just FYI, there are two Canada.ca blue colours:

- #26374a `#26374a` for buttons
-  `#284162` for link text

It's a Very Cool Feature of WET, as far as I can tell.

wordpress/wp-content/themes/cds-default/functions.php
```diff
@@ -87,6 +91,59 @@ function cds_setup(): void
...
+            array(
+                'name'  => __('Green', 'cds-snc'),
+                'slug'  => 'green',
+                'color' => '#00703c',
+            ),
+            array(
+                'name'  => __('Red', 'cds-snc'),
+                'slug'  => 'red',
+                'color' => '#af3c43',
+            ),
```
pcraig3, line 145: Did you grab these colours from anywhere or are they the defaults for something?

pcraig3 (approved):
Approving because this looks good + easy to change. I had some questions about the colours, so feel free to answer and then merge.

wordpress/wp-content/themes/cds-default/functions.php
timarney: Grabbed from the line under the heading.

wordpress/wp-content/themes/cds-default/functions.php
timarney: We should add the other one for sure + come up with better names and add missing colours to the palette  --- going to merge this one but for sure want to circle back.

There's also another way to select buttons which might be coming from the Bootstrap blocks plugin (not sure).  I would like the user to edit as need be and or pick a solid default.

Screen Shot 2021-11-01 at 2 41 52 PM

PR #212, opened 2021-11-01
Speed up cypress tests by removing screenshots and recording

dsamojlenko:
# Summary | Résumé

This seems to shave about a couple minutes off our cypress run - will make local runs a bit less annoying as well without all the screenshots.

Also caught a small type-o from a previous pr

### DIFF
--- a/wordpress/wp-content/mu-plugins/cds-base/classes/Modules/Users/Users.php
+++ b/wordpress/wp-content/mu-plugins/cds-base/classes/Modules/Users/Users.php
@@ -62,7 +62,7 @@ class Users
     public function sanitizeEmailAndRole($data = []): array|false
     {
         if (!is_array($data) && !is_object($data)) {
-            throw new ValidationException([ this->getEmailErrors(), $this->getRoleErrors() ]);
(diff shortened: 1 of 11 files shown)


wordpress/wp-content/mu-plugins/cds-base/classes/Modules/Users/Users.php
```diff
@@ -62,7 +62,7 @@ public function getRoles()
     public function sanitizeEmailAndRole($data = []): array|false
     {
         if (!is_array($data) && !is_object($data)) {
-            throw new ValidationException([ this->getEmailErrors(), $this->getRoleErrors() ]);
+            throw new ValidationException([ $this->getEmailErrors(), $this->getRoleErrors() ]);
```
dsamojlenko, line 65: type-o

pcraig3: opps, good catch

pcraig3 (approved):
All working fine for me 👍

PR #216, opened 2021-11-03
Don't modify ###SITENAME### for emails

timarney:
Digging into https://github.com/cds-snc/gc-articles-issues/issues/10 

User.php in WP core does a string replacement for "###SITENAME### 

Screen Shot 2021-11-03 at 8 44 14 AM

We're seeing this type of string in password resets etc...

```
This notice confirms that your password was changed on ###CollectionNAME###.
```

Which make sense as that's not going to get string replaced

**Note:**
> I haven't been able to test this locally but if we do a staging release believe this is worth a try. Locally I'm hitting WPML errors (S3 WPML issue) and a Guzzle error with Notify neither of which would be caused by this update.


pcraig3 (approved):
Makes sense, seems like it will work.

PR #217, opened 2021-11-03
Deactivate user accounts after 90 days

pcraig3:
# Summary

This PR adds the [Disable Users Login](https://wordpress.org/plugins/disable-user-login/) plugin and extends its functionality. It adds a daily job that will deactivate users who haven't logged in for 90 days.

1. Disable Users Login: allows admins to manually disable and enable user accounts
2. Adds `UserLockout` class, which schedules a daily job that deactivates users who haven't logged in for a while, and it adds a "login" whenever a disabled user is enabled. 

To test this PR, I would recommend changing the `USER_LOCKOUT_TIME` to something really short (maybe 60 seconds), and then adding `$this->setLoginLock();` to the construct method so that you can manually test the logic. You should see non-superadmins who haven't logged in for 60 seconds show up as disabled. You can then re-enable them and you should see entries for them show up in the `wp_userlogins` table.  

Users who are never disabled:
- superadmins
- users who have never logged in

Make sure to install the new plugin by doing `cd wordpress` & `composer update`.

## 1. Disable Users Login plugin

This plugin is pretty robust, adding most of what we need for disabling users:
- Add a function to prevent logging in + a customizable error message 
- Make "deactivated" users visible
- Create an edit field to re-activate a deactivated user (on the profile page, and in the table as well)

Here's a before and after, looking at the users table.

| before | after |
|--------|-------|
|  normal users table we see today      |    new column "disabled", and new "bulk actions" (you can't see them in the screenshot)   |
|  Screen Shot 2021-11-04 at 10 44 05   |    Screen Shot 2021-11-04 at 13 06 42   |

There's also a new setting in a user's profile (near the bottom):
 
Screen Shot 2021-11-04 at 10 43 03

The plugin looks like it's pretty good to work with as well:

- All the code is in [one heavily-commented PHP class](https://github.com/saintsystems/disable-user-login/blob/master/includes/class-ss-disable-user-login-plugin.php)
- [It allows you to add your own language file to translate the text](https://github.com/saintsystems/disable-user-login/blob/master/includes/class-ss-disable-user-login-plugin.php#L135-L148)
- I opened a PR and it was merged (https://github.com/saintsystems/disable-user-login/pull/5), so we can upstream changes if they make sense

So basically that's all good then.

## 2.  Adds `UserLockout` class

The UserLockout class checks if the Disable Users Plugin is active, and then if it is, it adds a scheduled job and a couple of hooks. If the plugin is inactive, it removes any tasks that may still be scheduled. 

There are 2 main functions here: 

1. Every day, check if there are users who haven't logged in for 90 days. If any are found, deactivate them. (Note: doesn't apply to superadmins or users who have never logged in).
2. If a `disabled` user is `enabled`, add a login for that user and use the hook name as the user agent. This prevents a situation where someone hasn't logged in for 90 days, they are re-enabled by their manager, but they don't log in before the end of the day and are locked out again.

I've also added a few extra methods to the `TrackLogins` class: one that lets us return the most recent login for a user, and one that lets us insert a login row. 

This class is kind of hard to test, especially the scheduled method, but it doesn't actually do too much and it cleans up after itself if the plugin it depends on goes away. So I think we're good here.

<details>
<summary>Background</summary>

There are a few ways to deactivate users in WordPress, but you end up having a lot of interface to think about. Let's say we want to set a meta field on users (`deactivated`), and then flip it to `true` after 90 days. Then we have to:

- Add a function to prevent logging in 
- Add a message for the user who is deactivated (do they alert their manager? do they know who they are?)
- Make "deactivated" users visible in the users table
- Create an edit field to re-activate a deactivated user (on the profile page, and maybe in the table as well)

It ends up being a reasonable end-to-end feature, so I started looking around for plugins that would handle this. The one I ended up using is called [Disable User Login](https://wordpress.org/plugins/disable-user-login/), which adds the ability for admins to manually disable anyone they want, and handles all the interface interactions.

Screen Recording 2021-11-03 at 10 33 04

## What's in this PR

Since we get all that for free, I wrote a class that schedules a job where it looks at the user's last login time and will deactivate them (using the same meta key as the plugin (`_is_disabled`)), depending on how much time has passed since their last login.

This PR does the following:

- Every _45 seconds_*, check all users
- Get their last login time
  - If never logged in before, they are exempt (can't deactivate a user before they have ever logged in)
- If their last login time is more then _60 seconds_* in the past, flip the `_is_disabled` switch on them

 *_Note_: these are values that make it easy to test the code actually runs. In reality, we want to run this once a day and check for 89 or 90 days since last login time. 

These users can never be locked:
- Super admins (I think the cron job doesn't have the permissions to do it)
- Users who have never logged in
</details>

### DIFF
--- a/wordpress/.snyk
+++ b/wordpress/.snyk
@@ -17,4 +17,9 @@ ignore:
         reason: None Given
         expires: 9999-12-30T00:00:00.000Z
         created: 2021-11-01T21:19:45.920Z
+  'snyk:lic:composer:wpackagist-plugin:disable-user-login:GPL-3.0':
+    - '*':
+        reason: None Given
+        expires: 9999-12-30T00:00:00.000Z
+        created: 2021-11-04T14:51:05.876Z
 patch: {}
--- a/wordpress/composer.json
+++ b/wordpress/composer.json
@@ -25,6 +25,7 @@
     "wpackagist-plugin/wp-bootstrap-blocks":"3.3.2",
     "wpml/sitepress-multilingual-cms": "4.5.0",
     "wpackagist-plugin/login-lockdown":"1.8.1",
+    "wpackagist-plugin/disable-user-login":"1.3.2",
     "wpackagist-plugin/wordpress-importer":"0.6.4",
     "wpackagist-plugin/wordpress-seo": "^16.9",
     "humanmade/s3-uploads": "^3.0"
@@ -44,7 +45,8 @@
         "wpackagist-plugin/wp-native-php-sessions",
         "wpackagist-plugin/wp-bootstrap-blocks",
         "humanmade/s3-uploads",
-        "wpackagist-plugin/login-lockdown"
+        "wpackagist-plugin/login-lockdown",
+        "wpackagist-plugin/disable-user-login"
       ]
     }
   },
--- a/wordpress/composer.lock
+++ b/wordpress/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "c1ebbbbadfc1b438a7c594fac55f425b",
+    "content-hash": "ed610cd90db31de27ac55be01a5571af",
     "packages": [
         {
             "name": "aws/aws-crt-php",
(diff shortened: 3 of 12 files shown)


timarney (approved):
👍

PR #219, opened 2021-11-03
Add user email text update

timarney:
Updates Add user to collection email text.

Closes https://github.com/cds-snc/gc-articles-issues/issues/50

See: https://github.com/cds-snc/gc-articles-issues/issues/50#issuecomment-956453978

### DIFF
--- a/wordpress/wp-content/mu-plugins/cds-base/classes/Modules/Users/Users.php
+++ b/wordpress/wp-content/mu-plugins/cds-base/classes/Modules/Users/Users.php
@@ -150,11 +150,11 @@ class Users
             'login'
         );
 
-        $subject  = "Set Password";
-        $message  = __('Someone requested that the password be reset for the following account:') . "\r\n\r\n";
-        $message .= __('If this was a mistake, just ignore this email and nothing will happen.') . "\r\n\r\n";
-        $message .= __('To set your password, visit the following address:') . "\r\n\r\n";
-        $message .=  $uniqueUrl;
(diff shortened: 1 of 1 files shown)


wordpress/wp-content/mu-plugins/cds-base/classes/Modules/Users/Users.php
```diff
@@ -150,11 +150,11 @@ public function sendReset($uId, $email)
             'login'
         );
 
-        $subject  = "Set Password";
-        $message  = __('Someone requested that the password be reset for the following account:') . "\r\n\r\n";
-        $message .= __('If this was a mistake, just ignore this email and nothing will happen.') . "\r\n\r\n";
-        $message .= __('To set your password, visit the following address:') . "\r\n\r\n";
-        $message .=  $uniqueUrl;
+        $subject = __("Invitation to collaborate on GC Articles", "cds-snc");
+        $message = __('Someone has invited this email to collaborate on a GC Articles collection site.', "cds-snc") . "\r\n\r\n";
```
pcraig3, line 154: Do we know who is sending them the email? If so, we could replace the "Someone". 

Although, seems like this works with or without it.

pcraig3 (approved):

wordpress/wp-content/mu-plugins/cds-base/classes/Modules/Users/Users.php
timarney: We should have the email of the person logged in but I'll skip that in this PR.

PR #220, opened 2021-11-03
Invite user - Text updates

timarney:
Closes - https://github.com/cds-snc/gc-articles-issues/issues/80

### DIFF
--- a/wordpress/wp-content/mu-plugins/cds-base/classes/Modules/Users/Users.php
+++ b/wordpress/wp-content/mu-plugins/cds-base/classes/Modules/Users/Users.php
@@ -98,7 +98,7 @@ class Users
         }
 
         if (!EmailDomains::isAllowedDomain(sanitize_email($email))) {
-            $error['message'] = __("You can’t use this email domain for registration.");
+            $error['message'] = __("You must enter Government of Canada email to send an invitation.");
             return $error;
         }
 
@@ -190,7 +190,7 @@ class Users
             return new WP_REST_Response([
                 [
                     "status" => $statusCode,
-                    "message" => $email . __(" was added to the Collection."),
+                    "message" => $email . __(" was invited to the Collection."),
                     "uID" => $uId,
                     "email" => $email
                 ]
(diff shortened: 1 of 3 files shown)


wordpress/wp-content/mu-plugins/cds-base/classes/Modules/Users/Users.php
```diff
@@ -98,7 +98,7 @@ private function getEmailErrors($email)
         }
 
         if (!EmailDomains::isAllowedDomain(sanitize_email($email))) {
-            $error['message'] = __("You can’t use this email domain for registration.");
+            $error['message'] = __("You must enter Government of Canada email to send an invitation.");
             return $error;
```
pcraig3, line 102: This should probably say "You must enter *a* Government of Canada email...".

Note that this error message is used for both of these cases:
- badly-formatted email addresses: anything that would fail a generic email validation (eg, `hello`)
- properly-formatted email addresses with a non-GC domain (`hello@canada.com`)

wordpress/wp-content/mu-plugins/cds-base/classes/Modules/Users/Users.php
```diff
@@ -190,7 +190,7 @@ public function addUserToCollection($data): WP_REST_Response|false
             return new WP_REST_Response([
                 [
                     "status" => $statusCode,
-                    "message" => $email . __(" was added to the Collection."),
+                    "message" => $email . __(" was invited to the Collection."),
```
pcraig3, line 193: This is a good catch. 

Although in the users table, they appear right away. Maybe we should add a flag somewhere that they are still just "invited".

pcraig3 (commented):
Just one comment. Looks like you missed word.

wordpress/wp-content/mu-plugins/cds-base/classes/Modules/Users/Users.php
timarney: Ya - it's not accurate as far as the users table goes + real status as the account is added but makes sense from a user entering the data end.

wordpress/wp-content/mu-plugins/cds-base/classes/Modules/Users/Users.php
timarney: 👍

wordpress/wp-content/mu-plugins/cds-base/classes/Modules/Users/Users.php
timarney: Would could add a flag (in another PR) to see if they've ever logged in ... starts to move into an update after your PR is merged https://github.com/cds-snc/gc-articles/pull/217

i.e. check for never logged in.

PR #225, opened 2021-11-04
Subscribe form updates

timarney:
Closes https://github.com/cds-snc/gc-articles-issues/issues/73

- Updates the subscribe form to allow users to enter a list ID via a dropdown
Also adds attributes for form labels (email field + button)

Screen Shot 2021-11-04 at 12 02 14 PM

### DIFF
--- a/wordpress/wp-content/mu-plugins/cds-base/classes/Modules/Subscribe/Block.php
+++ b/wordpress/wp-content/mu-plugins/cds-base/classes/Modules/Subscribe/Block.php
@@ -30,8 +30,16 @@ class Block
                     ],
                     'listId' => [
                         'type' => 'string',
-                        "default" => "0c188973-efab-4e8e-8b09-b165e98c66cf"
-                    ]
+                        "default" => ""
+                    ],
+                    'emailLabel' => [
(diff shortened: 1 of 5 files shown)


pcraig3 (commented):
Trying this locally, it mostly works but I don't see the List ID field.

-  I initially didn't see the new fields (had to `npm build` from the `classes/Modules/Subscribe` folder)
- Once I did that, I could see the others but not List ID

Screen Shot 2021-11-08 at 10 27 35

Do I need to change my .env file or anything, or should this just work?

timarney:
@pcraig3 


In your console do you have values for
```
CDS_VARS.notify_list_ids
```

??

Screen Shot 2021-11-08 at 1 22 45 PM

pcraig3 (changes requested):
Just one small change here: the `for` label attribute should match the `id` of the input element: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label#using_the_for_attribute

So you should add `id="email"` to the input and then we should be good to go.

pcraig3 (approved):
Approved! Feel free to merge when the tests pass.

PR #231, opened 2021-11-09
Add repeater settings interface

timarney:
In support of Part 5 here - https://github.com/cds-snc/gc-articles-issues/issues/69#issuecomment-962096516

Adds a settings repeater to handle more complex data entry for example where the user needs to enter JSON values

```
[{"id":"123", "type":"email", "label":"my-list"}]
```

This splits the entries into form entry fields for each value.

Screen Shot 2021-11-09 at 7 46 29 AM

The resulting values are saved to a hidden field in the existing format to keep things backwards compatible.

### DIFF
--- a/wordpress/wp-content/mu-plugins/cds-base/classes/Modules/Notify/NotifySettings.php
+++ b/wordpress/wp-content/mu-plugins/cds-base/classes/Modules/Notify/NotifySettings.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace CDS\Modules\Notify;
 
 use CDS\Modules\EncryptedOption\EncryptedOption;
+use InvalidArgumentException;
 
 class NotifySettings
 {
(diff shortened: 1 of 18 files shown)


wordpress/wp-content/mu-plugins/cds-base/classes/Modules/Notify/NotifySettings.php
```diff
@@ -279,14 +286,56 @@ public function listManagerApiKeyCallback()
...
+        } catch (Exception $exception) {
+            throw new InvalidArgumentException($exception->getMessage());
+        }
+    }
+
     public function listManagerNotifyServicesCallback()
     {
-        if ($string = $this->LIST_MANAGER_NOTIFY_SERVICES) {
-            $this->getObfuscatedOutputLabel($string, 'list_manager_notify_services_value');
+        $serviceIdData = get_option('LIST_MANAGER_NOTIFY_SERVICES');
```
timarney, line 313: Move this into try catch

dsamojlenko (approved):

PR #233, opened 2021-11-09
Give GC Admins ability to see and manage categories

pcraig3:
# Summary

Super simple PR: adds "manage categories" permission to all GC Admins, and bumps the DB version so that roles are cleaned up with their new permissions.

Resolves this issue: https://github.com/cds-snc/gc-articles-issues/issues/40

And also, interestingly, GC Admins don't see the line of text about tags referenced in this issue:  https://github.com/cds-snc/gc-articles-issues/issues/44
(superadmins can still see it but we don't care about them)


timarney:
Just need to fix up the conflict

timarney (approved):

PR #242, opened 2021-11-15
Hooks to prevent new users being created on a new network site.

pcraig3:
# Summary

This is a proof of concept PR with some working code that:
- _says_ you can only add Super Admins to a new collection 
- in reality, prevents _new users_ from being created when creating a new collection

You can still assign a non-super admin if you want — it's kind of arbitrary because all super admins are able to make changes to new sites after the fact. It would be nicer to actually restrict this to super users, but this is the best you can do without replacing the page.

super-users

## Explanation

The code that creates a new network site is here: https://github.com/WordPress/WordPress/blob/a207dadc10e88208a82e6b94f204b0122df97d95/wp-admin/network/site-new.php#L36

It's a mix of functions and HTML code, so it's hard to extract anything in isolation without replacing the whole page. There are **3** hooks on this page, and no filters:

1. _Before_ adding a new user
2. _After_ adding a new user
3. _Before_ rendering the "submit" button (but _after_ the rest of the form has already been rendered)

- Using the 1st hook, we can prevent new users from being added (by using `wp_die` — returning a `WP_Error` is ignored).
- Using the 3rd hook, we can add more HTML to the page. In this case, I used CSS to hide the previous message about creating new users and I added new message saying that only Super Admins are accepted.

Unfortunately, there is no way to override parts of this form (for example, removing the "Admin Email" field and replacing it with a `<select>`), or the underlying logic for checking users (we can only stop sites being created for new users, we can't check the role of existing users).

As far as I can tell, it's either do something like this or replace the page.

### DIFF
--- a/wordpress/wp-content/mu-plugins/cds-base/classes/Modules/Cleanup/SitesToCollections.php
+++ b/wordpress/wp-content/mu-plugins/cds-base/classes/Modules/Cleanup/SitesToCollections.php
@@ -6,6 +6,9 @@ namespace CDS\Modules\Cleanup;
 
 class SitesToCollections
 {
+    public string $adminEmailMessage;
+    public string $collectionMessage;
+
     public function __construct()
     {
         if (is_admin()) {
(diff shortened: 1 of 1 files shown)


timarney:
Reviewed --- we'll run with this type of setup.

timarney:
This looks good -- up for debate where the code belongs ... the original intent for `SitesToCollections.php` was for string replacements from Site(s) to Collection(s).

Thoughts?

pcraig3:
I don't really mind where it goes. I figured it should probably live in `Cleanup` since we're cleaning up some loose ends, and then I was looking around for where we have code that's related to "Collections" as a concept. I could create a new class called "Collections" if you don't like it in here. 

(I also thought of renaming the class from `SitesToCollections`, but the extra verboseness seems like it's helpful in signalling what it does.)

timarney (approved):

pcraig3:
So create a new class or merge as-is?

timarney:
merge as-is it's fine.

PR #246, opened 2021-11-15
Catch when the notify services option is empty (new site)

dsamojlenko:
# Summary | Résumé

On new sites, when the LIST_MANAGER_NOTIFY_SERVICES option doesn't exist yet, this was erroring out. ~~Moved the deserialize call into the try/catch to catch that problem~~.

After some digging, decided to change how the underlying methods work - instead of throwing an exception on empty or invalid input, they methods just return an empty array, which is what the try/catch blocks were doing anyway.

### DIFF
--- a/wordpress/wp-content/mu-plugins/cds-base/classes/Modules/Notify/FormHelpers.php
+++ b/wordpress/wp-content/mu-plugins/cds-base/classes/Modules/Notify/FormHelpers.php
@@ -81,6 +81,8 @@ class FormHelpers
 
           <?php
             $sender = new NotifyTemplateSender();
+            $services = [];
+
             $serviceIdData = get_option('LIST_MANAGER_NOTIFY_SERVICES');
             $services = Utils::deserializeServiceIds($serviceIdData);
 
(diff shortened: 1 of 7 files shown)


timarney (approved):

pcraig3 (changes requested):
I'm still getting this error on the "Send Notify Template" page 

```
PHP Fatal error:  
Uncaught InvalidArgumentException: 
No service data in /usr/src/gc-articles/wordpress/wp-content/mu-plugins/cds-base/classes/Modules/Notify/Utils.php:15
\nStack trace:
\n#0 /usr/src/gc-articles/wordpress/wp-content/mu-plugins/cds-base/classes/Modules/Notify/FormHelpers.php(85): CDS\\Modules\\Notify\\Utils::deserializeServiceIds('')
\n#1 /usr/src/gc-articles/wordpress/wp-content/mu-plugins/cds-base/classes/Modules/Notify/NotifyTemplateSender.php(241)
```

Looks like it's coming from [`FormHelpers.php`](https://github.com/cds-snc/gc-articles/blob/d3c926cebf6e364ee99ca450ceed737ea040b135/wordpress/wp-content/mu-plugins/cds-base/classes/Modules/Notify/FormHelpers.php#L85)

pcraig3 (approved):
Approved! Looks good, works for me now. 👍

PR #250, opened 2021-11-16
Add bulk send cap profile checkbox

timarney:
Adds a checkbox for bulk sending permission under a users profile.

Screen Shot 2021-11-16 at 10 07 07 AM

Closes https://github.com/cds-snc/gc-articles-issues/issues/23

### DIFF
--- a/wordpress/wp-content/mu-plugins/cds-base/classes/Modules/Cleanup/Roles.php
+++ b/wordpress/wp-content/mu-plugins/cds-base/classes/Modules/Cleanup/Roles.php
@@ -41,7 +41,7 @@ class Roles
                 'manage_network_users' => 1, // enables "edit_users" for GC Admins'
                 'manage_options' => 0,
                 'manage_notify' => 1,
-                'manage_list_manager' => 1,  // by default this is off 
+                'manage_list_manager' => 1,  // by default this is off
                 'list_manager_bulk_send' => 0, // this is managed per user (see user profile)
                 'manage_categories' => 1,
                 'read' => 1,
--- a/wordpress/wp-content/mu-plugins/cds-base/classes/Modules/Cleanup/Roles.php
+++ b/wordpress/wp-content/mu-plugins/cds-base/classes/Modules/Cleanup/Roles.php
@@ -8,8 +8,7 @@ class Roles
 {
     public function __construct()
     {
-        Utils::checkOptionCallback('cds_base_activated', '1.0.11', function () {
-
+        Utils::checkOptionCallback('cds_base_activated', '1.0.15', function () {
             if (is_blog_installed()) {
                 remove_role('editor');
                 remove_role('author');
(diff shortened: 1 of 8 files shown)


wordpress/wp-content/mu-plugins/cds-base/classes/Modules/Cleanup/Roles.php
```diff
@@ -8,8 +8,7 @@ class Roles
 {
     public function __construct()
     {
-        Utils::checkOptionCallback('cds_base_activated', '1.0.11', function () {
-
+        Utils::checkOptionCallback('cds_base_activated', '1.0.15', function () {
```
pcraig3, line 11: Could be `1.0.12`, but it doesn't really matter.

wordpress/wp-content/mu-plugins/cds-base/classes/Modules/Notify/ListManagerUserProfile.php
```diff
@@ -0,0 +1,68 @@
...
+
+    public function __construct()
+    {
+    }
+
+    public static function register()
+    {
+        $instance = new self();
+
+        add_action('show_user_profile', [$instance, 'displayListManagerMeta']);
```
pcraig3, line 18: this one only shows up for superadmins, right? It seems like we don't really need it, since we can't turn this off if we're superadmins (it is just checked again if you reload the page a couple times).

wordpress/wp-content/mu-plugins/cds-base/classes/Modules/Notify/ListManagerUserProfile.php
```diff
@@ -0,0 +1,68 @@
...
+        add_action('edit_user_profile', [$instance, 'displayListManagerMeta']);
```
pcraig3, line 19: We can also add this permissions to GC Editors. That's fine, right? They still aren't able to see Notify settings, but they do get the List Manager page.

pcraig3 (commented):
Just a few comments, very small overall.

wordpress/wp-content/mu-plugins/cds-base/classes/Modules/Notify/ListManagerUserProfile.php
timarney: Yes - we'll allow a super admin to add a GC Editor to send.

wordpress/wp-content/mu-plugins/cds-base/classes/Modules/Cleanup/Roles.php
timarney: Will leave as is --- I was testing locally a bunch of times.

wordpress/wp-content/mu-plugins/cds-base/classes/Modules/Notify/ListManagerUserProfile.php
timarney: So yes the Admin would setup the Notify settings as need be.

wordpress/wp-content/mu-plugins/cds-base/classes/Modules/Notify/ListManagerUserProfile.php
timarney: Updated

pcraig3 (approved):
👍

PR #259, opened 2021-11-17
Update change sites label to search for strings that contain ###

timarney:
Closes: https://github.com/cds-snc/gc-articles-issues/issues/10

Updating `changeSitesLabel` to use string contains 

The issue with the previous PR was the string being passed in != `the text` because the string contains a sentence vs a word. 

```... This notice confirms that your password was changed on ###SITENAME### more text ...```

Ref: https://github.com/cds-snc/gc-articles/pull/216
Ref: https://github.com/cds-snc/gc-articles/pull/61

Note: Ultimately we need to find a more efficient way to fix these translations but this will fix the current implementation

### Testing 

> Ensure you have your Notify API key etc... setup i.e. you are receiving WP emails

On the current `main` branch 

1) On your local installation edit a user 
```
http://localhost/wp-admin/user-edit.php?user_id=YOUR_USER_ID
```

2) Click on the `reset password` button --- no further action needed after the click
Screen Shot 2021-11-17 at 8 21 44 AM

3)  Click `update user`
Screen Shot 2021-11-17 at 8 22 12 AM

4) Wait for the email from Notify email 
Screen Shot 2021-11-17 at 8 25 47 AM

Not the string replacement is failing.

<hr>

Switch to the `chore/10` branch

... repeat the steps above and note the string now gets replaced.

### DIFF
--- a/wordpress/wp-content/mu-plugins/cds-base/classes/Modules/Cleanup/SitesToCollections.php
+++ b/wordpress/wp-content/mu-plugins/cds-base/classes/Modules/Cleanup/SitesToCollections.php
@@ -24,7 +24,7 @@ class SitesToCollections
             return $translation;
         }
 
-        if (str_contains($translation, '###')) { 
+        if (str_contains($translation, '###')) {
             return $translation;
         }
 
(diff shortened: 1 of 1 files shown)


pcraig3 (approved):
Looks good!

Screen Shot 2021-11-17 at 11 57 06

I am feeling the love from all at CDS wordpress base.

PR #262, opened 2021-11-17
Re-add the "Network" admin menu

pcraig3:
# Summary 

I don't know why it's missing but I am pretty sure we want it back. Adds back the "My Collections" (ie, "my sites") menu into the top menu bar for superadmins.

Note that if you're a regular GC Admin, you can be signed up to multiple sites, but you don't see the "My Collections" option in the top menu. This means we will eventually want a way of routing people between the sites they are a part of, but for now we can re-add this option without other roles being able to see it.

## Screenshot

| before | after |
|--------|-------|
|  Screen Shot 2021-11-17 at 12 13 54    |   Screen Shot 2021-11-17 at 12 13 26  |


timarney:
Let's add this back only for super admins

For GC Admins we introduced the My Collections panel on the dashboard to serve this purpose.
Screen Shot 2021-11-17 at 1 05 40 PM

timarney (approved):

PR #264, opened 2021-11-18
Navigation - Add gutenburg plugin + fix options ul styling

timarney:
Adds the Gutenburg plugin to "unlock" the navigation block which is currently in `beta`

Closes https://github.com/cds-snc/gc-articles-issues/issues/117

## Example
 
Screen Shot 2021-11-18 at 7 48 02 AM

Screen Shot 2021-11-18 at 7 48 16 AM

 ## Setup
 
 Not I opted not to auto activate the plugin for now and also it's not added as a MU Plugin so we can control what sites we enable it for.  You can make it available under network settings.

Screen Shot 2021-11-18 at 7 51 20 AM

Screen Shot 2021-11-18 at 7 51 49 AM

 ## Note
 
 Behind the scenes the navigation block ends up adding data just like the existing Appearance -> Menu setup so we don't need to worry about the block being in `beta`. 
 
Screen Shot 2021-11-18 at 8 01 04 AM

### DIFF
--- a/wordpress/composer.json
+++ b/wordpress/composer.json
@@ -28,7 +28,8 @@
     "wpackagist-plugin/disable-user-login":"1.3.2",
     "wpackagist-plugin/wordpress-importer":"0.6.4",
     "wpackagist-plugin/wordpress-seo": "^16.9",
-    "humanmade/s3-uploads": "^3.0"
+    "humanmade/s3-uploads": "^3.0",
+    "wpackagist-plugin/gutenberg":"11.9.1"
   },
   "require-dev": {
     "pestphp/pest": "^1.16",
--- a/wordpress/composer.lock
+++ b/wordpress/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "8bf05fd18cbb89ca1b9325a080afdaf3",
+    "content-hash": "16b910c4ca83566a7a0408731aaa01de",
     "packages": [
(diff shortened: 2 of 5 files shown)


timarney:
@pcraig3 -- I did a minor patch to fix the UL styles for the sidebar in this PR as well given the user will need to interact more with that panel to set the menu location etc... 

Screen Shot 2021-11-18 at 7 59 49 AM

pcraig3:
I installed / activated the plugin, and then clicked the checkbox for the experimental menu, so that's all good. However, when I press buttons on the new menu I created, I just see a bunch of errors show up. Can you think of anything I'm doing wrong?

Screen Recording 2021-11-18 at 12 33 40

timarney:
@pcraig3 -- I'm seeing those issues as well locally when creating a new menu.  That said I can create a new menu using "the old way" and edit it using the block.  


Given we're not "counting" on this plugin i.e. it doesn't block us at this point I'm going to suggest we put this in ... setup the base menus to edit and go from there (that would at least get us "edit mode").  Taking this approach also gets us started with something to investigate with as far as our setup -- spin it up with wp-env etc...  Wondering if any of our other plugins are interfering.

timarney:
Looks like it's an issue outside for "all" wp install --- tested with a plain version of WP Env. 

Still game to start with this --- and I imagine they'll fix that in the next release.   I'll see if I can find an issue on the Gutenburg  repo

Screen Shot 2021-11-18 at 1 45 19 PM

pcraig3 (approved):
Yup, I'm good to merge this as well.

I will continue using the old menu system for now but keep an eye on this one.

PR #265, opened 2021-11-18
Rough in header and footer menus

timarney:
Registers - header, footer and side nav area

Checks to see if header and footer menus have content --- if not they will use the current Canada.ca defaults.

cc: @pcraig3 

## Testing

Add a menu and assign it to the header or footer location - Appearance -> Menus

Note:
I have left out the side nav which will need to be added to index.php, page.php etc...
I've also left the menu output fully unstyled i.e. this is prepping for the next sprint


pcraig3 (approved):
Works! Looks great also 😉

PR #268, opened 2021-11-18
Force hide menus

timarney:
Adding a variable to hide the header and footer menus for now.


pcraig3 (approved):
I don't know why, I think it looks amazing.

Screen Shot 2021-11-18 at 15 54 09

PR #272, opened 2021-11-19
Add media access for GC Admins and GC Editor + force grid view

timarney:
closes https://github.com/cds-snc/gc-articles-issues/issues/96

This PR updates the permissions to allow GC Admins and GC Editor users to upload media.

I've also 
- updated the Media list to force "Grid" view
- remove filter options

### DIFF
--- a/wordpress/wp-content/mu-plugins/cds-base/classes/Modules/Cleanup/Media.php
+++ b/wordpress/wp-content/mu-plugins/cds-base/classes/Modules/Cleanup/Media.php
@@ -31,10 +31,6 @@ class Media
     {
         ?>
         <style type="text/css">
-        .view-switch .view-list {
-            display: none;
-        }
-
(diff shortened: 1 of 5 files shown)


wordpress/wp-content/mu-plugins/cds-base/classes/Modules/Cleanup/Media.php
```diff
@@ -0,0 +1,58 @@
...
+
+add_action('admin_head', function () {
+    ?>
+    <style type="text/css">
+        .view-switch .view-list {
+            display: none;
+        }
+    <style>
+    <?php
+});
```
pcraig3, line 58: Don't need this: it is already accounted for on line 35.

pcraig3 (approved):
Looks good! Just one comment about some code you can remove.

PR #273, opened 2021-11-19
Add customizable header menu and footer menu

pcraig3:
# Summary | Résumé

Hello hello,

This PR gives us some initial menu cusomize-ability and makes some small quality of life tweaks. 

In this PR:
- If you create a header menu, it will overwrite the blue "Menu" bar entirely and replace it with our grey bar.
- If you create a footer menu, it will replace the grey links in the footer with your new links.
- Fixes the "top" link
- Removes the "home" class from the homepage (causes a conflict with WET)
- Fixes for the menu button

**Not in this PR**:

- Add a side menu (still working it out)
- Add drop-downs for the top menu (TBD)
- Add a specific "mobile" menu collapsible element (like [the Notify homepage](https://notification.canada.ca/), as an example)

### Header 

| Admin interface | frontend |
|--------|-------|
|    Screen Shot 2021-11-19 at 09 40 39  |    Screen Shot 2021-11-19 at 09 41 12   ​|

### Footer

| Admin interface | frontend |
|--------|-------|
|   Screen Shot 2021-11-19 at 09 34 21    |   Screen Shot 2021-11-19 at 09 34 28  |

### The top link 

You can only see this on mobile, but the link wasn't actually pointing anywhere, now it points to the `<header>` element.

Screen Shot 2021-11-22 at 10 04 24

### The menu button

The "Canada.ca" menu is _only_ white on _one page_ on Canada.ca: [the homepage](https://www.canada.ca/en.html). On every other page, it's blue. On the Canada.ca homepage, they use `class="home"` on the body element to change the colour. It just so happens that WordPress also uses the same class be default, and so it was making our menu white (probably doing some other stuff as well, but I didn't really check). I changed the classname to "homepage" instead and then the menu bar turned blue.

I also noticed that the blues being used for the menu bar vs the blue line were not the same, so I fixed it, and I squared off the top edges of the menu button.

| before | after |
|--------|-------|
|   two-different blues, weird rounded top-corners     |  same blue, squared-off at the top     |
|  Screen Shot 2021-11-22 at 10 05 34    |   Screen Shot 2021-11-22 at 10 05 55   |

<details>

<summary>Questions for the team (answered by Tim)</summary>

## Questions for the team

- Is it okay making these modifications directly in the theme header/footer/styles files, rather than the plugin?
- Currently it doesn't handle nested items. I was thinking we only want to allow 1 level of nesting at the top, and 0 on the bottom. Is this right? or should we just have no levels of nesting everywhere. 
- There's no "mobile" style for the menu ([here's an example (make your screen small)](https://alpha.service.canada.ca/home)), do we want one if there are lots of items?

The side menu I haven't implemented yet. Should it be a modification to the basic layout file (force people into a 2/3rds layout), or is it something we want to have available as a block and then people can drag and drop it in the gutenberg editor?

<details>


timarney:
> Is it okay making these modifications directly in the theme header/footer/styles files, rather than the plugin?

Yes -- these are specific to the theme.  

> Currently it doesn't handle nested items. I was thinking we only want to allow 1 level of nesting at the top, and 0 on the bottom. Is this right? or should we just have no levels of nesting everywhere.

We'll cap at one level for now.  Moving forward I would like to support 2 levels but happy to start with 1.  We won't cap in the admin just let folks know as needed for now.

> There's no "mobile" style for the menu (here's an example (make your screen small)), do we want one if there are lots of items?

Yes we'll want mobile --- if you want to do that in a separate PR we can do that.

pcraig3:
> > Is it okay making these modifications directly in the theme header/footer/styles files, rather than the plugin?

> Yes -- these are specific to the theme.

All good then

> > Currently it doesn't handle nested items. I was thinking we only want to allow 1 level of nesting at the top, and 0 on the bottom. Is this right? or should we just have no levels of nesting everywhere.

> We'll cap at one level for now. Moving forward I would like to support 2 levels but happy to start with 1. We won't cap in the admin just let folks know as needed for now.

Okay, for this PR it's just a flat menu structure but we can expand it later.

> > There's no "mobile" style for the menu (here's an example (make your screen small)), do we want one if there are lots of items?

> Yes we'll want mobile --- if you want to do that in a separate PR we can do that.

Sounds good, separate PR it is.

> > Should it be a modification to the basic layout file (force people into a 2/3rds layout),

> For now let's force the 2/3rds layout if someone setup up the left nav.

This is more more than I anticipated, so I've started another branch for it locally and I'm working on it separately.

wordpress/wp-content/themes/cds-default/footer.php
```diff
@@ -3,18 +3,59 @@
...
+        <div class="container">
+            <div class="row">
+                <nav class="col-md-9 col-lg-10 ftr-urlt-lnk">
+                    <h2 class="wb-inv">About this site</h2>
+                    <ul>
+                    <?php
+                        $locations = get_nav_menu_locations();
+
+                        // Get menu assigned to location, or false
+                        $footerMenu = wp_get_nav_menu_object($locations['footer']);
```
timarney, line 30: Need to add an isset here 

isset($locations['footer'])

timarney:
Received this error when I first swapped over to this branch

Screen Shot 2021-11-22 at 11 16 25 AM

wordpress/wp-content/themes/cds-default/footer.php
```diff
@@ -3,18 +3,59 @@
...
-if ($footerMenu && $showMenu) {
-    echo $footerMenu;
-} else {
-    require_once 'footer_' . $lang . '.php';
-}
+<div style="display:none;" id="version" style="margin-top:30px;"><?php echo _S_VERSION; ?></div>
+<footer id="wb-info">
+    <div class="landscape">
+        <nav class="container wb-navcurr">
+            <h2 class="wb-inv">About government</h2>
```
timarney, line 14: Needs translation wrapper

wordpress/wp-content/themes/cds-default/footer.php
```diff
@@ -3,18 +3,59 @@
...
+            <ul class="list-unstyled colcount-sm-2 colcount-md-3">
+                <?php echo print_menu_links($footerText['aboutGovernment']); ?>
+            </ul>
+        </nav>
+    </div>
+    <div class="brand">
+        <div class="container">
+            <div class="row">
+                <nav class="col-md-9 col-lg-10 ftr-urlt-lnk">
+                    <h2 class="wb-inv">About this site</h2>
```
timarney, line 24: Needs translation wrapper?

timarney (approved):

PR #275, opened 2021-11-19
Hide "tag" references

timarney:
Quick PR to hide reference to tags. 

Not the best way I'm sure to do this but given the # of times this page is likely to be visited I think it's reasonable way to go.

closes https://github.com/cds-snc/gc-articles-issues/issues/44
closes https://github.com/cds-snc/gc-articles-issues/issues/42


pcraig3 (approved):
Looks good 👍

PR #281, opened 2021-11-23
Update login title

timarney:
Closes 

https://github.com/cds-snc/gc-articles-issues/issues/104

Replaces `WordPress` with `GC Articles`
Screen Shot 2021-11-23 at 8 53 30 AM


pcraig3 (approved):
👍  Works, looks good

PR #282, opened 2021-11-23
Force "all" for category link

timarney:
Forces "all" languages url for category link click
https://github.com/cds-snc/gc-articles-issues/issues/43

Note: I tried to do this using PHP but looks like WPML does a redirect to EN to get around that using jQuery

## Testing
- Click categories from the left nav of the WP Admin --- it should default to "all" languages.


pcraig3 (approved):
👍

PR #290, opened 2021-11-24
Add homepage pattern

timarney:
Adds a new block pattern "Homepage"

Screen Shot 2021-11-24 at 2 22 14 PM

### DIFF
--- a/wordpress/wp-content/themes/cds-default/functions.php
+++ b/wordpress/wp-content/themes/cds-default/functions.php
@@ -190,7 +190,7 @@ function print_menu_links(array $links): string
         $link = (object)$link; // cast to object so that we can use arrow notation
         $string .= sprintf(
             "<li><a href='%s'>%s</a></li>",
-            clean_url($link->url),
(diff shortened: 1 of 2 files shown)


wordpress/wp-content/themes/cds-default/functions.php
```diff
@@ -190,10 +190,79 @@ function print_menu_links(array $links): string
         $link = (object)$link; // cast to object so that we can use arrow notation
         $string .= sprintf(
             "<li><a href='%s'>%s</a></li>",
-            clean_url($link->url),
+            esc_url($link->url),
```
pcraig3, line 193: Yes, I noticed this as well.

pcraig3 (approved):
Looks good!

It seemed a little weird to me that the H1 size (`1.9em`) is so close to the H2 size (`1.8em`), but this appears to be the standard WET thing to do ([example page](https://www.canada.ca/en/immigration-refugees-citizenship/services/coronavirus-covid19/visitors.html))

PR #291, opened 2021-11-24
Clean up the side menu a bit

pcraig3:
# Summary | Résumé

This PR is about refactoring the codebase so that the "menu building" functions can be centralized in the "template-functions" file rather than just cluttering up the page templates.

- Move footer links function into `template-functions`
- Move side nav function into `template-functions`. User `ob_start` so that we can nest the HTML more easily (hard to see when it's a big string)
- Swap out use of `clean_url` (deprecated) for `esc_url` (not deprecated)
- Create cypress function to add 2 pages, assign the second one to be the child of the first, and then check there is a menu.

No functional / design changes are introduced by this PR.

If the cypress test is too much, I can put it on ice. It was a little tricky to get working, so it might turn out to be brittle in the future.

### DIFF
--- a/wordpress/wp-content/themes/cds-default/footer.php
+++ b/wordpress/wp-content/themes/cds-default/footer.php
@@ -14,7 +14,7 @@ $footerMenu = '';
         <nav class="container wb-navcurr">
             <h2 class="wb-inv"><?php echo $footerText['aboutGovernmentTitle']; ?></h2>
             <ul class="list-unstyled colcount-sm-2 colcount-md-3">
-                <?php echo print_menu_links($footerText['aboutGovernment']); ?>
+                <?php echo print_footer_links($footerText['aboutGovernment']); ?>
             </ul>
         </nav>
     </div>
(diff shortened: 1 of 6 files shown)


wordpress/wp-content/themes/cds-default/footer.php
```diff
@@ -14,7 +14,7 @@
         <nav class="container wb-navcurr">
             <h2 class="wb-inv"><?php echo $footerText['aboutGovernmentTitle']; ?></h2>
             <ul class="list-unstyled colcount-sm-2 colcount-md-3">
-                <?php echo print_menu_links($footerText['aboutGovernment']); ?>
+                <?php echo print_footer_links($footerText['aboutGovernment']); ?>
```
timarney, line 17: I know this is an existing function name now but seems a little odd to say "echo print" maybe move the echo inside the function as we'll know it's meant to print.

timarney:
Looks good - just the one comment which is code that already exists and the conflict to fix up.

wordpress/wp-content/themes/cds-default/footer.php
pcraig3: Maybe `get_footer_links` is more reasonable? 
Or some of the PHP-y methods are like `wp_header()`, so this could just be `footer_links()`?

timarney (approved):

pcraig3:
Going to merge!

PR #292, opened 2021-11-25
Add send email to existing user being added to a collection

timarney:
Closes https://github.com/cds-snc/gc-articles-issues/issues/126

### DIFF
--- a/wordpress/wp-content/mu-plugins/cds-base/classes/Modules/Users/Users.php
+++ b/wordpress/wp-content/mu-plugins/cds-base/classes/Modules/Users/Users.php
@@ -189,14 +189,12 @@ class Users
         $userInfo = get_userdata($uId);
 
         $blogName = get_bloginfo('name');
+        $url = get_bloginfo('url');
 
         // phpcs:disable
         $subject = __("Invitation to collaborate on GC Articles", "cds-snc"). " — ". $blogName;
-        
-        $blogInfo->name;
-        
-        $message = __('Someone has invited this email to collaborate on a GC Articles collection site.', "cds-snc") . "\r\n\r\n";
-        
+        $message .= __('Someone has invited this email to collaborate on a GC Articles — ', "cds-snc").$blogName. "\r\n\r\n";
+        $message .= __('To get started log in here:', "cds-snc").$url;
         
         // phpcs:enable
 
--- a/wordpress/wp-content/mu-plugins/cds-base/classes/Modules/Users/Users.php
+++ b/wordpress/wp-content/mu-plugins/cds-base/classes/Modules/Users/Users.php
@@ -190,7 +190,7 @@ class Users
 
         $blogName = get_bloginfo('name');
         $url = network_site_url('wp-login.php');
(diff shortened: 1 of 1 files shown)


wordpress/wp-content/mu-plugins/cds-base/classes/Modules/Users/Users.php
```diff
@@ -184,6 +184,25 @@ public function sendReset($uId, $email)
...
+    public function sendAddToCollection($uId, $email)
+    {
+        $userInfo = get_userdata($uId);
+
+        $blogName = get_bloginfo('name');
+
+        // phpcs:disable
+        $subject = __("Invitation to collaborate on GC Articles", "cds-snc"). " — ". $blogName;
+        
+        $blogInfo->name;
```
pcraig3, line 196: This throws an error (and it's not used for anything anyway. It feels like it was left in by accident.)

wordpress/wp-content/mu-plugins/cds-base/classes/Modules/Users/Users.php
```diff
@@ -184,6 +184,25 @@ public function sendReset($uId, $email)
...
+        
+        $message = __('Someone has invited this email to collaborate on a GC Articles collection site.', "cds-snc") . "\r\n\r\n";
```
pcraig3, line 198: It would be good to have a link to the site as well. 

I created a new collection to test this, so I got both of the emails one after the other.

| creating a new site | adding a pre-existing user |
|---------------------|----------------------------|
|     Lots of info in here: new site url, your username, url for login page, two bullet points (🤔 )                |      Almost no info in here. Would be good to have  the same 2 links as the other one: link to site, link to login page.  We can skip the username since it's the same as the email address.                 |
|    Screen Shot 2021-11-25 at 11 35 51           |  Screen Shot 2021-11-25 at 11 39 08 |

The one thing not pictured is that the site name is in the subject line for both, which is good. But getting the URLs would still be part of making this actionable.

pcraig3 (changes requested):
Couple of comments!

wordpress/wp-content/mu-plugins/cds-base/classes/Modules/Users/Users.php
timarney: Added a url + the site name to the messaging.

wordpress/wp-content/mu-plugins/cds-base/classes/Modules/Users/Users.php
timarney: removed

pcraig3 (approved):
Approved! 

Here's the screenshot of the email I'm getting, for reference.


Screen Shot 2021-11-25 at 13 35 10

We're MVP all day every day, here so all good. But if we wanted to make it nicer for the future, I would suggest: 
- the first sentence is a bit weird: "a GC Articles"
- Who invited me?

All to say: good for now and if we are doing polishing in the future, we can look at this.

PR #297, opened 2021-11-26
Hide user count

timarney:
Closes https://github.com/cds-snc/gc-articles-issues/issues/125


pcraig3 (approved):
Works! Let's hope nobody needs to sort between their 2 types of users.

PR #300, opened 2021-11-26
Fix side nav styles with really specific CSS

pcraig3:
# Summary | Résumé

It looks like the CSS file for gutenberg block library (`block-library/style.css?ver=11.9.1`) got updated, so it started forcing columns to be 50% width even for our side nav, which is using the same classes as the default wordpress column classes. The solution? Make our CSS more specific and now it works. This is the kind of thing that would look a lot better in SCSS but w/e, we can live with it.

### FAQ

> Why are we using wordpress classes if they conflict with our own CSS?

Basically, it's because I want to keep the markup as similar as possible to what you would get using Gutenberg. It's kind of a weak rationale, but there it is.

## Screenshots 

| before | after |
|--------|-------|
|   Screen Shot 2021-11-26 at 13 25 20   |  Screen Shot 2021-11-26 at 13 25 41   |

### DIFF
--- a/wordpress/wp-content/themes/cds-default/style.css
+++ b/wordpress/wp-content/themes/cds-default/style.css
@@ -192,24 +192,35 @@ nav.nav--about li a:focus {
     }
 }
 
-/* Side nav layout */
-@media screen and (min-width: 768px) {
-    .page__side-nav {
-        flex-basis: 35%;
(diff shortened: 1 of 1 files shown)


timarney:
> block-library/style.css?ver=11.9.1

That version matches the Gutenburg plugin

timarney (approved):

PR #301, opened 2021-11-26
Fix en_US 404

timarney:
Tracked down the code that sets the url path for i18n

Screen Shot 2021-11-26 at 2 22 14 PM

The code uses document.documentElement.lang 

Which was returning `en-US`

Screen Shot 2021-11-26 at 2 21 13 PM

Canada.ca -- sets the val as `en`.

This PR updates the lang attribute and in turn fixes the request.


pcraig3 (approved):
Awesome! Good catch 👍

PR #302, opened 2021-11-26
Upgrade to Yoast Premium

dsamojlenko:
# Summary | Résumé

Upgrades to Yoast Premium. To install Yoast Premium, we needed to add a composer auth credential. Previously we were adding a Github token for installing the WPML plugin from our private mirror. Rather than re-serializing the json and updating that secret, took a slightly different approach generating the auth.json file with composer commands.

Also uncommented a block related to NPM cache in our Cypress github action file. This was supposed to be done with #289 

## To install
You will need a new environment variable to install from the private Yoast repository. You can retrieve this from LastPass and then set it in your environment file:

```
COMPOSER_YOAST_TOKEN=[TOKEN]
```

Then you can run
- `docker-compose run composer` to install the new dependencies, then 
- `docker-compose run install` to activate them.

Alternatively, you can run `composer install` from the wordpress folder. In that case, you will be prompted for a username and password. Username is `token` and password is the token that you can retrieve from LastPass.

Note that either way you do it, you will see a big block of warnings at the end of the composer install. According to Yoast Support, this is a known issue, and should not affect the plugin.

image


timarney (approved):

timarney:
Merge this when your set @dsamojlenko

PR #309, opened 2021-11-29
Drop-down + mobile top nav

pcraig3:
# Summary | Résumé

Adding a top nav with a drop-down and a mobile "Menu" button. Also added some aria-labels because I ran an axe scan and it found some small things.
- Top nav now supports 1 level of drop-down (not 2)
- Top nav will collapse on mobile
- Removed the "landmarks must be unique" axe warning
- Removed the "empty header" axe warning

### Drop-down menu

The drop-down menu markup was taken from [CSS tricks](https://css-tricks.com/solved-with-css-dropdown-menus/). The idea was that we don't want to create a drop-down that's just a button, where you can't click the link anymore. It also suggests using an `aria-label`, so I added one.

- On large screens, it opens when you hover or focus the parent element.
- On small screens, we just list all the links, so it doesn't collapse.

### Mobile nav

The mobile nav is an "open/close" pattern that I copied from [the Bootstrap Navbar](https://getbootstrap.com/docs/5.1/components/navbar/). I've copied what bootstrap does with aria-roles, which is pretty minimal.

## gifs

Screen Recording 2021-11-30 at 11 33 43

### DIFF
--- a/wordpress/wp-content/themes/cds-default/inc/template-functions.php
+++ b/wordpress/wp-content/themes/cds-default/inc/template-functions.php
@@ -400,8 +400,10 @@ function get_top_nav(): string
 
         // Insert aria-label for top nav (otherwise axe complains)
         $dom->find('.nav--primary__container')->setAttribute('aria-label', $topMenu);
-        // // Insert aria-label for submenu
+        // Insert aria-label for submenu
         $dom->find('.sub-menu')->setAttribute('aria-label', $submenu);
+        // Insert aria-expanded for link with submenu
+        $dom->find('.menu-item-has-children')->setAttribute('aria-expanded', "false");
 
         return $dom->outerHTML;
     }
--- a/wordpress/wp-content/themes/cds-default/inc/template-functions.php
+++ b/wordpress/wp-content/themes/cds-default/inc/template-functions.php
@@ -164,9 +164,9 @@ function cds_breadcrumb($sep = ''): string
         $html = $child->firstChild()->innerHtml;
         $parts = explode('|', $html);
 
-        $output = '<nav id="wb-bc" property="breadcrumb">';
+        $output = '<nav id="wb-bc" property="breadcrumb" aria-label="' .  __("Breadcrumbs") . '">';
         $output .= '<div class="container">';
-        $output .= '<h2><?php _e("You are here:"); ?></h2>';
+        $output .= '<h2>' . __("You are here:") . '</h2>';
         $output .= '<ol class="breadcrumb">';
         // note this will need to point to the correct language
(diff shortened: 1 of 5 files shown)


timarney:
Maybe for another PR but what do you think about adding `aria-expanded="true"` when the dropdowns are open .  Currently when I tested using VoiceOver I didn't get an indication.

This is a pretty good post on the topic

http://www.ashleysheridan.co.uk/blog/Making+an+Accessible+Hamburger+Menu

wordpress/wp-content/themes/cds-default/inc/template-functions.php
```diff
@@ -164,9 +164,9 @@ function cds_breadcrumb($sep = ''): string
         $html = $child->firstChild()->innerHtml;
         $parts = explode('|', $html);
 
-        $output = '<nav id="wb-bc" property="breadcrumb">';
+        $output = '<nav id="wb-bc" property="breadcrumb" aria-label="' .  __("Breadcrumbs") . '">';
         $output .= '<div class="container">';
-        $output .= '<h2><?php _e("You are here:"); ?></h2>';
+        $output .= '<h2>' . __("You are here:") . '</h2>';
```
timarney, line 169: Need to add domain here `__("You are here:", 'cds-snc')`

wordpress/wp-content/themes/cds-default/inc/template-functions.php
```diff
@@ -164,9 +164,9 @@ function cds_breadcrumb($sep = ''): string
         $html = $child->firstChild()->innerHtml;
         $parts = explode('|', $html);
 
-        $output = '<nav id="wb-bc" property="breadcrumb">';
+        $output = '<nav id="wb-bc" property="breadcrumb" aria-label="' .  __("Breadcrumbs") . '">';
```
timarney, line 167: `__("Breadcrumbs", 'cds-snc')`

wordpress/wp-content/themes/cds-default/inc/template-functions.php
```diff
@@ -362,3 +362,51 @@ function get_side_nav(): string
...
+        "echo" => false,
+        "depth" => 2,
+        "menu_class" => "nav nav--primary container",
+        "menu_id" => $menuID,
+        "container" => "nav",
+        "container_class" => "nav--primary__container"
+    ]);
+
+    if ($headerMenu) {
+        $topMenu = __('Top menu');
```
timarney, line 391: __('Top menu', 'cds-snc');

timarney (approved):

pcraig3:
Boom 💥

PR #310, opened 2021-11-30
Update which table user logins read and write to

timarney:
Closes https://github.com/cds-snc/gc-articles-issues/issues/122

Updates the table used to read and write table logins 

We want to use the global vs the site tables

see: https://paulund.co.uk/get-database-table-prefix-in-wordpress

### DIFF
--- a/wordpress/wp-content/plugins/cds-base/classes/Modules/TrackLogins/TrackLogins.php
+++ b/wordpress/wp-content/plugins/cds-base/classes/Modules/TrackLogins/TrackLogins.php
@@ -17,7 +17,7 @@ class TrackLogins
         global $wpdb;
         $this->wpdb      = $wpdb;
 
-        if ($wpdb->base_prefix) {
+        if (property_exists($wpdb, 'base_prefix')) {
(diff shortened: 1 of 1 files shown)


wordpress/wp-content/plugins/cds-base/classes/Modules/TrackLogins/TrackLogins.php
```diff
@@ -16,7 +16,12 @@ public function __construct()
     {
         global $wpdb;
         $this->wpdb      = $wpdb;
-        $this->tableName = $this->wpdb->prefix . 'userlogins';
+
+        if (property_exists($wpdb, 'base_prefix')) {
+            $this->tableName = $wpdb->base_prefix . 'userlogins';
+        } else {
+            $this->tableName = $this->wpdb->prefix . 'userlogins';
+        }
```
dsamojlenko, line 24: Is there a chance base_prefix doesn't exist?

Also, should we add something to cleanup/remove the existing tables? (maybe they don't matter as we won't migrate them to prod and eventually we'll wipe staging)

dsamojlenko (approved):

PR #311, opened 2021-11-30
Add a screenshot for the theme

timarney:
Adding a Maple Leaf for a theme icon / screenshot  --- this replaces the "empty" .png.   We should the the only ones to see this but nice to have something show up so the theme doesn't look broken. 

This is the same Maple Leaf used for the `preview` but it filled in not just an outline.

Screen Shot 2021-11-30 at 12 22 21 PM


pcraig3 (approved):
🍁

PR #312, opened 2021-11-30
Add css fix -- for live preview menu button

timarney:
Turns off the live preview button .
Screen Shot 2021-11-30 at 1 35 35 PM

Think this is a bad idea?

Checkout the WP source code :)

wp-admin/css/themes.css
Screen Shot 2021-11-30 at 1 31 22 PM


pcraig3 (approved):
lol @ the classname.

Does that affect more than just this button? It seems like they might have used in in lots of places for specific themes (like ours) that don't want the customizer menu to be visible.

timarney:
Yes -- but I think because we don't want to show the customizer this is fine overall.

pcraig3:
Yeah, the whole "customize" menu option is hidden now.

Screen Shot 2021-11-30 at 17 11 33

PR #314, opened 2021-12-01
Add new Apache container and config

dsamojlenko:
# Summary | Résumé

Prepares an Apache container and pushes to our new container repository (#313). This simply builds the new container and pushes it to ECR, it will not affect any existing services.

In preparation for #305


pcraig3 (approved):
👍

PR #320, opened 2021-12-01
Fixes for the top menu

pcraig3:
# Summary | Résumé

There's 2 changes in this PR:
1. Fix the problem where the whole site disappears if there are no drop-down items on a custom menu
2. Fix the problem where nested submenus would get their own chevron

### The whole site disappears

WordPress returns menus as a big HTML string that you are supposed to `echo` out on the page, so you can only modify the HTML once you have the big string. I am using [PHP HTML Parser](https://github.com/paquettg/php-html-parser) to find elements in the menu and then add aria attributes.

However, I was finding the element, assuming it existed, and then trying to add attributes. When the element didn't exist (in our case, no submenu) PHP would error out and the page wouldn't be rendered.

| before | after |
|--------|-------|
|  nothing rendered to the page after the FIP      |   all good    |
| Screen Shot 2021-12-01 at 11 21 12  |   Screen Shot 2021-12-01 at 11 21 43  |

### Nested submenus would get their own chevron

Technically, the WordPress menu builder lets you create menus to an arbitrary depth, but we are limiting people to a depth of `2`.  However, the menu returned still adds the `has-children` class to leaf nodes that think they have a submenu (but they don't). 

Our CSS was adding chevrons to these second-level menu items, but we want that to go away.

| before | after |
|--------|-------|
|   `AUTHOR'S NOTE` has a chevron     |  `AUTHOR'S NOTE` _does not_ have a chevron     |
|   Screen Shot 2021-12-01 at 11 20 36   |  Screen Shot 2021-12-01 at 11 20 05 |

### DIFF
--- a/wordpress/wp-content/themes/cds-default/inc/template-functions.php
+++ b/wordpress/wp-content/themes/cds-default/inc/template-functions.php
@@ -388,32 +388,37 @@ function get_top_nav(): string
     ]);
 
     if ($headerMenu) {
-        $topMenu = __('Top menu', 'cds-snc');
-        $submenu = __('submenu', 'cds-snc');
-
         // Insert a button (markup taken from bootstrap)
         // It seems like we can't append an element using the PHP HTML Parser https://stackoverflow.com/q/51466367
         $headerMenu = str_replace('<nav class="nav--primary__container">', '<nav class="nav--primary__container"><div class="container"><button class="navbar-toggler" type="button" data-toggle="collapse" data-target="#' . $menuID . '" aria-controls="' . $menuID . '" aria-expanded="false">Menu</button></div>', $headerMenu);
 
-        $dom = new Dom();
-        $dom->loadStr($headerMenu);
+        try {
+            $topMenu = __('Top menu', 'cds-snc');
+            $submenu = __('submenu', 'cds-snc');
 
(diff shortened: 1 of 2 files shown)


wordpress/wp-content/themes/cds-default/inc/template-functions.php
```diff
@@ -388,24 +388,37 @@ function get_top_nav(): string
...
+
+            // Insert aria-expanded for link with submenu
+            $menuItemNode = $dom->find('.menu-item-has-children')[0];
+            if ($menuItemNode) {
+                $menuItemNode->setAttribute('aria-expanded', "false");
+            }
+
+            return $dom->outerHTML;
+        } catch (Exception) {
+            return $headerMenu;
```
timarney, line 420: Will this return "unstyled" if it reaches here?

timarney (approved):

wordpress/wp-content/themes/cds-default/inc/template-functions.php
pcraig3: Nope, it looks the same as before. all the CSS still applies and I'm adding the menu button outside of this try/catch block so the mobile menu still works as well.

PR #360, opened 2021-12-03
Make "expander" element true to life

pcraig3:
# Summary | Résumé

Added better markup and the WET CSS for the expander element in the block editor itself, removed the `h3`, and also made the previous "summary" block into an "Inner Block" so that you can just add different things.

Expanders can now house all the other blocks:
- headings
- lists
- images
- etc

The only annoying thing is that they will open and close during editing, which I don't know how to stop. 😬 

Note: make sure to do an npm build before testing this.

## Screenshots

| before | after |
|--------|-------|
|  hacking it together      |  can use multiple blocks  (lists, images, headings, etc)   |
|   Screen Shot 2021-12-03 at 16 57 35    |  Screen Shot 2021-12-03 at 15 02 59   |

Once it's actually published, it looks like how it looked in the editor.

Screen Shot 2021-12-03 at 15 03 18

### DIFF
--- a/wordpress/wp-content/plugins/cds-base/classes/Modules/Blocks/src/editor-styles/editor.css
+++ b/wordpress/wp-content/plugins/cds-base/classes/Modules/Blocks/src/editor-styles/editor.css
@@ -3,6 +3,8 @@
 .editor-styles-wrapper p,
 .editor-styles-wrapper ol, 
 .editor-styles-wrapper ul,
+.editor-styles-wrapper details,
+.editor-styles-wrapper details *,
 .wp-block-table table,
 .wp-block-table figcaption {
   font-family: 'Noto Sans', sans-serif;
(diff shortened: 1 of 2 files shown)


timarney:
The updated expander makes sense and works well --- we'll run into issues for anyone using the current expander as the markup has changed and Gutenburg won't understand how to parse / upgrade to the new format.

See: https://www.youtube.com/watch?v=m456H94sAj4

For this we have 2 options

1) Look into how to create deprecations in turn making the update possible without causing issues
 https://developer.wordpress.org/block-editor/reference-guides/block-api/block-deprecation/
 
 2) Create a new block called accordion and we'll get folks to "upgrade"  / replace existing content.  
 
 For sure either way we should look into  `deprecations` for moving forward as we'll need to modify blocks in the future.

PR #362, opened 2021-12-06
Add fields to contact form

timarney:
Initial PR to add contact form fields when requesting a site.

I'll create another PR to only allow adding this block if your a `super admin`
Screen Shot 2021-12-06 at 1 02 27 PM

### DIFF
--- a/wordpress/wp-content/plugins/cds-base/classes/Modules/Contact/ContactForm.php
+++ b/wordpress/wp-content/plugins/cds-base/classes/Modules/Contact/ContactForm.php
@@ -43,7 +43,6 @@ class ContactForm
         ?>
         <div class="gc-form-wrapper">
             <?php if (isset($_POST['contact-type'])) {
-
                 $contactType = $_POST['contact-type'];
                 $heading = '';
                 $label = '';
@@ -81,8 +80,7 @@ class ContactForm
                 }
                 ?>
 
-                <?php if ($contactType === 'request-site'): ?>
-                    
+                <?php if ($contactType === 'request-site') : ?>
                     <form id="contact-form" method="POST" action="/wp-json/contact/v1/process">
                     <input type="hidden" name="contact-type" value="<?php echo $contactType; ?>"/> 
                     <h2><?php echo $heading; ?></h2>
@@ -174,7 +172,7 @@ class ContactForm
                     </div>
                 </form>
                 
-                <?php else: ?>
+                <?php else : ?>
                 <form id="contact-form" method="POST" action="/wp-json/contact/v1/process">
                     <h2><?php echo $heading; ?></h2>
                     <?php wp_nonce_field(
@@ -209,7 +207,7 @@ class ContactForm
                 </form>
 
                 <?php endif; ?>
-               <?php
+                <?php
             } else {
                 $current_url = home_url(add_query_arg([], $wp->request)); ?>
                 <form id="contact-form-step-1" method="POST" action="<?php echo $current_url; ?>">
@@ -218,7 +216,7 @@ class ContactForm
                             <span id="contact_types" class="hidden"><?php _e(
                                 'Contact Types',
                                 'cds-snc',
-                            ); ?></span>
+                                                                    ); ?></span>
                         </label>
                         <div class="focus-group" style="margin-bottom: 1.75rem">
                             <?php $this->radioField(
--- a/wordpress/wp-content/plugins/cds-base/classes/Modules/Contact/ContactForm.php
+++ b/wordpress/wp-content/plugins/cds-base/classes/Modules/Contact/ContactForm.php
@@ -213,10 +213,7 @@ class ContactForm
                 <form id="contact-form-step-1" method="POST" action="<?php echo $current_url; ?>">
                     <div role="group" aria-labelledby="contact_types">
                         <label for="contact_types">
-                            <span id="contact_types" class="hidden"><?php _e(
-                                'Contact Types',
-                                'cds-snc',
-                            ); ?></span>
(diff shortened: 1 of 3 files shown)


wordpress/wp-content/plugins/cds-base/classes/Modules/Contact/ContactForm.php
```diff
@@ -42,45 +42,147 @@ public function render($atts, $content = null): string
...
+                    <div class="focus-group">
+                    <label class="gc-label required" id="collection-name-label" for="gc-collection-name">
+                        <?php _e("What do you want to name your GC Articles collection?", "cds-snc"); ?>
+                    </label>
+                    <input
+                            type="text"
+                            class="gc-input-text"
+                            id="gc-collection-name"
+                            required
+                            aria-describedby="collection-name-label"
```
pcraig3, line 111: I'm pretty sure you don't need `aria-describedby` when we already have the label pointing to this.

wordpress/wp-content/plugins/cds-base/classes/Modules/Contact/ContactForm.php
```diff
@@ -42,45 +42,147 @@ public function render($atts, $content = null): string
...
+                            placeholder=""
+                            name="gc-collection-name"
+                    ></input>
+                    </div>
+
+                    <!-- heard about  -->
+                    <div class="focus-group">
+                    <label class="gc-label required" id="heard-about-from=label" for="heard-about-from">
```
pcraig3, line 119: `heard-about-from=label` has an unintended equals sign -- looks like you meant `heard-about-from-label`

wordpress/wp-content/plugins/cds-base/classes/Modules/Contact/ContactForm.php
```diff
@@ -42,45 +42,147 @@ public function render($atts, $content = null): string
...
+                        <?php _e("How did you hear about GC Articles?", "cds-snc"); ?>
+                    </label>
+                    <input
+                            type="text"
+                            class="gc-input-text"
+                            id="heard-about-from"
+                            required
+                            aria-describedby="heard-about-from-label"
```
pcraig3, line 127: Yeah, here and the ones below all have `aria-describedby` as well, but I don't think it's needed.

wordpress/wp-content/plugins/cds-base/classes/Modules/Contact/ContactForm.php
```diff
@@ -42,45 +42,147 @@ public function render($atts, $content = null): string
...
+                            placeholder=""
+                            name="heard-about-from"
+                    ></input>
+                    </div>
+
+                     <!-- purpose  -->
+                     <div class="focus-group">
+                     <label class="gc-label required" id="purpose-label" for="purpose">
+                        <?php _e("What will you use your GC Articles collection for?
+", "cds-snc"); ?>
```
pcraig3, line 137: newline here is probably unintentional

dsamojlenko (approved):

wordpress/wp-content/plugins/cds-base/classes/Modules/Contact/ContactForm.php
```diff
@@ -99,49 +201,64 @@ class="gc-textarea"
...
+                $current_url = home_url(add_query_arg([], $wp->request)); ?>
                 <form id="contact-form-step-1" method="POST" action="<?php echo $current_url; ?>">
                     <div role="group" aria-labelledby="contact_types">
                         <label for="contact_types">
-                            <span id="contact_types" class="hidden"><?php _e("Contact Types", "cds-snc") ?></span>
+                            <span id="contact_types" class="hidden"><?php _e('Contact Types', 'cds-snc'); ?></span>
                         </label>
                         <div class="focus-group" style="margin-bottom: 1.75rem">
-                            <?php $this->radioField("contact-type", "request-site", __("Request a site", "cds-snc")); ?>
                             <?php $this->radioField(
```
pcraig3, line 219: It's a little bit odd that there is initially no form heading, and then one shows up. 

eg, here's me clicking "request a site"

| before | after |
|--------|-------|
|  no heading above the form, so I added `Contact form` as an `<h3>`       |  now we have `Contact form` as a `<h3>` and `Request a site` as an `<h2>`      |
|   Screen Shot 2021-12-07 at 09 55 39    |   Screen Shot 2021-12-07 at 09 56 27    |


If we put this on a page with no other info -- just and `h1` and this contact form -- then it will be fine. But if it's halfway down the page, it might be a weird experience.

wordpress/wp-content/plugins/cds-base/classes/Modules/Contact/Setup.php
```diff
@@ -73,14 +73,47 @@ public function confirmSend(): string
...
+        } else {
+            if (
+                (!isset($_POST["message"]) || $_POST["message"] === "")
+                || (!isset($_POST['contact-type']) || $_POST['contact-type'] === "")
+                || (!isset($_POST['email']) || $_POST['email'] === "")
+            ) {
+                return json_encode(["error" => __("Please complete the required field to continue", "cds-snc")]);
+            }
+
+            return json_encode($this->sendEmail($_POST["email"], $message, $_POST["contact-type"]));
```
pcraig3, line 116: This should use `$_POST["message"]` instead of `$message`, which hasn't been set yet.

pcraig3 (changes requested):
Just a few small changes. The most important one is the "$message" variable in at then end of `Setup.php`, which causes a 500 error so that none of the forms using it will send an email.

PR #368, opened 2021-12-06
Wire up and use the origin access identity path

dsamojlenko:
# Summary | Résumé

Cloudfront wants the path not the arn for the origin


timarney (commented):
🥇

PR #371, opened 2021-12-07
Hide contact form for non super admin

timarney:
Follow up PR for:
https://github.com/cds-snc/gc-articles/pull/362

Only allow Super Admins to add a contact form

### DIFF
--- a/wordpress/wp-content/plugins/cds-base/classes/Modules/Contact/Block.php
+++ b/wordpress/wp-content/plugins/cds-base/classes/Modules/Contact/Block.php
@@ -20,10 +20,10 @@ class Block
 
     public function registerBlock()
     {
-        if(!is_super_admin() && !Utils::isWpEnv()){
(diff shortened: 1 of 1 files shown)


dsamojlenko (approved):

pcraig3 (commented):
Can we customize the message here?

If I am an admin, this is what I see:

Screen Shot 2021-12-07 at 10 26 48

It's telling me that I should remove this block because it isn't supported, but it clearly _is_ working, I just don't have permissions to access it.

timarney:
> Can we customize the message here?

No sure if we can customize that said --- I don't think we have any user / testers outside of ourselves that  dropped that in --- it's now an "internal form" vs a general contact form.

timarney:
Going to merge but will look into how we can customize this type of message as we might need to do that in the future.

pcraig3 (approved):
Sounds good 👍 

Yeah, since local admins can't use it, it doesn't matter.

PR #372, opened 2021-12-07
Update home url

timarney:
Closes https://github.com/cds-snc/gc-articles-issues/issues/152


pcraig3 (commented):
Is there a setting that needs to be set in the DB? I am trying this locally and I am still getting "Canada.ca" as the home URL.

pcraig3 (approved):
My mistake, I was clicking around on `https://articles.cdssandbox.xyz`, not at localhost :|

PR #373, opened 2021-12-07
Remove public access from S3 bucket

dsamojlenko:
# Summary | Résumé

Once we get the S3 Uploads config (#369) deployed to production, we can remove public access to the S3 bucket.

**Heads-up** Don't merge this until after our next deployment.


timarney (approved):

pcraig3 (approved):
👍

PR #374, opened 2021-12-07
Remove sync menu

timarney:
Closes 

https://github.com/cds-snc/gc-articles-issues/issues/150


pcraig3 (approved):
I was going to say, we should only do this for "GC Admins", but it seems like that's how it works already.

PR #380, opened 2021-12-07
Update progress bar colour

timarney:
Update the progress bar colour to increase contrast.

See https://github.com/cds-snc/gc-articles-issues/issues/154#issuecomment-988239436 for details


timarney:
Tested using this image which is the "max" file size upload.

SampleJPGImage_1mbmb

pcraig3 (approved):
👍

PR #398, opened 2021-12-10
Change port number from load balancer -> ecs

dsamojlenko:
# Summary | Résumé

Follow up to #396 because we can't bind to a different host port, we must run on 8443 from LB -> ECS


infrastructure/terragrunt/aws/network/security-groups.tf
```diff
@@ -13,7 +13,7 @@ resource "aws_security_group_rule" "wordpress_ecs_ingress_lb" {
   description              = "Security group rule for WordPress ingress from load balancer"
   type                     = "ingress"
   from_port                = 443
```
patheard, line 15: This one needs to be 8443 as well.

patheard (changes requested):
Looks good!  Just one more spot to change to `8443` and we should be good:
https://github.com/cds-snc/gc-articles/blob/main/infrastructure/terragrunt/aws/network/security-groups.tf#L72-L73

This one lets the `8443` traffic go from the LB to the ECS cluster in the private subnets.

timarney (approved):

patheard (approved):

PR #405, opened 2021-12-13
Fix errors when re-setting password

timarney:
Closes https://github.com/cds-snc/gc-articles-issues/issues/173

Looks like Yoast and WMPL have more values that are expected when the user profile is submitted.  Given we're removing those fields the profile was throwing an error.

### DIFF
--- a/wordpress/wp-content/plugins/cds-base/classes/Modules/Cleanup/Profile.php
+++ b/wordpress/wp-content/plugins/cds-base/classes/Modules/Cleanup/Profile.php
@@ -13,6 +13,7 @@ class Profile
         add_action('personal_options', [$this, 'start']);
         add_filter('additional_capabilities_display', [$this, 'removeAdditionalCapabilitiesFunc']);
         add_action('wpml_user_profile_options', [$this, 'wpmlOptions']);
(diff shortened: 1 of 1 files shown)


wordpress/wp-content/plugins/cds-base/classes/Modules/Cleanup/Profile.php
```diff
@@ -128,8 +128,30 @@ public function removeAdditionalCapabilitiesFunc(): bool
...
+
+        $fields = ["honorificPrefix",
+        "honorificSuffix",
+        "birthDate",
+        "gender",
+        "award",
+        "knowsAbout",
+        "knowsLanguage",
+        "jobTitle",
+        "worksFor"];
```
pcraig3, line 144: I can't believe we are taking away "award" after that cool TBS one we got for COVID Alert

pcraig3 (approved):
👍  changed my password, all good

wordpress/wp-content/plugins/cds-base/classes/Modules/Cleanup/Profile.php
timarney: It's still on for Super Admins @pcraig3 so you can still input your excellence for that field.

PR #408, opened 2021-12-13
Add exception for WAF managed rule set

dsamojlenko:
# Summary | Résumé

CloudFront started throwing a 403 error on the WPML installer path. This path is being rejected by a WAF managed ruleset, probably due to the `/` characters towards the end of the querystring

`https://articles.cdssandbox.xyz/daves-test/wp-admin/admin.php?page=sitepress-multilingual-cms/menu/setup.php`

This PR will add an exception to that rule for paths under `/wp-admin`

### DIFF
--- a/infrastructure/terragrunt/aws/load-balancer/waf.tf
+++ b/infrastructure/terragrunt/aws/load-balancer/waf.tf
@@ -206,8 +206,8 @@ resource "aws_wafv2_web_acl" "wordpress_waf" {
     name     = "Custom_LFI_QueryString"
     priority = 3
 
-    override_action {
-      none {}
+    action {
+      block {}
     }
 
     visibility_config {
@@ -249,8 +249,8 @@ resource "aws_wafv2_web_acl" "wordpress_waf" {
     name     = "Custom_AWSManagedRulesLinuxRuleSet"
     priority = 4
 
-    override_action {
-      none {}
+    action {
+      block {}
     }
 
     visibility_config {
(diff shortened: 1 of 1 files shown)


infrastructure/terragrunt/aws/load-balancer/waf.tf
```diff
@@ -50,6 +50,184 @@ resource "aws_wafv2_web_acl" "wordpress_waf" {
...
+    override_action {
+      none {}
+    }
+
+    statement {
+      managed_rule_group_statement {
+        name        = "AWSManagedRulesLinuxRuleSet"
+        vendor_name = "AWS"
+        excluded_rule {
+          name = "awswaf:managed:aws:linux-os:LFI_QueryString"
```
patheard, line 88: I think this should just be the rule's name `LFI_QUERYSTRING` like so:
```hcl
name = "LFI_QUERYSTRING"
```

infrastructure/terragrunt/aws/load-balancer/waf.tf
```diff
@@ -50,6 +50,184 @@ resource "aws_wafv2_web_acl" "wordpress_waf" {
...
+          not_statement {
+            statement {
+              byte_match_statement {
+                field_to_match {
+                  uri_path {}
+                }
+                positional_constraint = "CONTAINS"
+                search_string         = "/wp-admin"
+                text_transformation {
+                  type     = "NONE"
```
patheard, line 132: If you set this to `LOWERCASE` it'll cause the rule to behave more predictably since it'll catch paths like `/WP-ADMIN`

patheard (changes requested):
A few small tweaks and I'd say it's ready to go.  If this works, it's a nice pattern we'll be able to use on all our WAF ACLs.

patheard (approved):

PR #409, opened 2021-12-14
Pattern image update

timarney:


pcraig3 (approved):
holdon-toyourbutts

I think we should just throw caution to the wind here and merge it.

PR #417, opened 2021-12-15
Scale down Staging database

dsamojlenko:
# Summary | Résumé

Scales down the Staging database cluster, and parameterizes instance_class so we can have different settings for staging/production.

cds-snc/gc-articles-issues#116


patheard (approved):
Looks good!

Just a note that this instance class will take place during the next maintenance window tonight.

PR #418, opened 2021-12-15
Add ECS placement strategy

dsamojlenko:
# Summary | Résumé

Re: cds-snc/gc-articles-issues#3

Adds an ECS placement strategy set so that the ECS tasks are spread evenly across the Virtual Private Cloud (VPC) availability zones


patheard (approved):
I'm looking forward to see how this works 🤞🏻

PR #420, opened 2021-12-16
Add assets used by the project

timarney:
There are some assets that we should "store" in the repo i.e. the EN, FR images for WPML and the Maple Leaf which is used for the Site Icon.


pcraig3 (approved):
All good!

pcraig3:
Also, Blaze it 420 bro   🚬😎 ☮️ (PR number)

PR #426, opened 2021-12-22
Add Workflow to build and push Production container

dsamojlenko:
# Summary | Résumé

Adds a workflow to build and push a tagged container to our Production ECR when a release is created.

Main differences between this workflow and the staging workflow:
- Only runs on release:created
- Checks out the git repo at the tag created during the release
- Pushes to Production ECR (obviously)


pcraig3 (approved):
Sweet! Let's `Push containers to ECR`!

PR #430, opened 2021-12-23
Load env file for composer action

dsamojlenko:
# Summary | Résumé

Fixes a bug where the env file wasn't being loaded for the composer install action in docker-compose.


pcraig3 (approved):
Let's goooooo!

PR #432, opened 2021-12-29
Add ids for skip to content

timarney:
Adds missing ids for the rest of the template files

ref: https://github.com/cds-snc/gc-articles-issues/issues/181


pcraig3 (approved):
Just noting that Canada.ca sites put the focus on the `<h1>`, but I think we're okay here.

PR #433, opened 2021-12-29
Add maintenance page with settings toggle

timarney:
Adds a Collections Settings page:
 http://localhost/your_site/wp-admin/options-general.php?page=collection-settings

Screen Shot 2021-12-29 at 1 14 31 PM

The setting will toggle the "maintenance" page for the theme.

  https://github.com/cds-snc/gc-articles/pull/433/files#diff-bba02e6b920d4630bb05d7c96faf707707377049ee86facfd29338e952cba6caR266


wordpress/wp-content/plugins/cds-base/classes/Modules/Site/SiteSettings.php
```diff
@@ -0,0 +1,98 @@
...
+            'collection_settings_section', // section
+            [
+                'label_for' => 'collection_mode'
+            ]
+        );
+    }
+
+    public function collectionSettingsSectionInfo()
+    {
+    }
```
pcraig3, line 88: Is this needed?

wordpress/wp-content/plugins/cds-base/classes/Modules/Site/SiteSettings.php
```diff
@@ -0,0 +1,98 @@
...
+
+    public function collectionSettingsCreateAdminPage()
+    {
+
+        ?>
+
+        <div class="wrap">
+            <h1><?php _e('Collection Settings', 'cds-snc') ?></h1>
+            <p></p>
```
pcraig3, line 47: Empty paragraph, we can probably lose this.

wordpress/wp-content/themes/cds-default/header.php
```diff
@@ -101,7 +101,7 @@
 
   <?php
     $headerMenu = get_top_nav();
-    if ($headerMenu) :
+    if ($headerMenu && !cds_is_maintenance_mode()) :
```
pcraig3, line 104: I don't know if we care, but there is also a custom footer menu that might show up. For consistency, we could disallow both.

Screen Shot 2021-12-29 at 13 14 30

wordpress/wp-content/themes/cds-default/maintenance.php
```diff
@@ -0,0 +1,28 @@
...
+ *
+ * This is the most generic template file in a WordPress theme
+ * and one of the two required files for a theme (the other being style.css).
+ * It is used to display a page when nothing more specific matches a query.
+ * E.g., it puts together the home page when no home.php file exists.
+ *
+ * @link https://developer.wordpress.org/themes/basics/template-hierarchy/
+ *
+ * @package cds-default
+ */
```
pcraig3, line 14: We should probably change this comment.

wordpress/wp-content/plugins/cds-base/classes/Modules/Site/SiteSettings.php
```diff
@@ -0,0 +1,98 @@
+<?php
+
+declare(strict_types=1);
+
+namespace CDS\Modules\Site;
+
+use InvalidArgumentException;
```
pcraig3, line 7: Does this get used?  It looks like I can enter any value I want by editing the "value" in the HTML.

pcraig3 (changes requested):
Really simple and it works well 👍 

A few things I noticed:

## Saving values

### Two flash messages

Whenever I save either value (or nothing), I get two different "Settings saved." flash messages. Not sure why this is happening, but it looks wrong.

Screen Shot 2021-12-29 at 13 17 58

### Any value is accepted (even submitting without any checkboxes selected)

I can hack the HTML to put in any value, or I can submit with neither value selected and it always says "Settings saved". In practice, it seems like this doesn't matter because there is a strict comparison to decide whether maintenance mode is on or off, but maybe it's not what you would expect as a user?

## Publishing

### No index?

If someone is in maintenance mode, we should no-index their site (eg, [search for "ignore" on this page for an example](https://jec.fyi/blog/setting-up-seo-and-google-analytics)). Easy enough to do.

### Allowing people to preview?

You can still click "preview in new tab" and "publish", but you always get the maintanence page template. Maybe it would be better to not show the maintenance page:
- When previewing specifically (or)
- When someone is logged in

## Page design

It looks like there should be more on the maintanence page (maybe a "get in touch" email address or a contact form?), but we can wait for Brian or someone else to make that call.

wordpress/wp-content/plugins/cds-base/classes/Modules/Site/SiteSettings.php
timarney: Just tested --- will remove

wordpress/wp-content/plugins/cds-base/classes/Modules/Site/SiteSettings.php
timarney: removed

wordpress/wp-content/themes/cds-default/header.php
timarney: Updated

timarney:
> When someone is logged in

Added a check for this ---  the page will only show if in maintenance and not logged in now.

timarney:
> Page design

Page design --- yes ... it's a bit of combination page i.e. thinking of when a site is just getting setup vs down for true maintenance for sure we can circle back on this.

timarney:
@pcraig3 

> No index?

Maybe we can pull the Front Page + No index settings into this page as well

https://github.com/cds-snc/gc-articles-issues/issues/178

timarney:
@pcraig3 

I added a dropdown so you can select a Maintenance page to control the page content. 

Screen Shot 2021-12-30 at 10 08 54 AM

wordpress/wp-content/plugins/cds-base/classes/Modules/Site/SiteSettings.php
```diff
@@ -0,0 +1,124 @@
...
+            array(
+                'name'              => 'collection_mode_maintenance_page',
+                'echo'              => 1,
+                'show_option_none'  => __('&mdash; Select &mdash;'),
+                'option_none_value' => '0',
+                'selected'          => get_option('collection_mode_maintenance_page'),
+            )
+        );
+    }
+}
```
pcraig3, line 124: This is way smarter than trying to do something really customizable. (ie, add a bunch of weird options to the settings page ("Page title", "Page content", etc).

pcraig3 (approved):
Just a couple of comments, but I am going to pre-approve.

1. `noindex` if in maintenance mode

Yeah, we can pull this into the other issue: https://github.com/cds-snc/gc-articles-issues/issues/178 . Let's leave a note in there.

2. Page title for maintenance page is site default

Just trying it out, it uses the site default page title instead of the title of the maintenance page. We should probably do the latter.

Screen Shot 2022-01-04 at 18 04 33

timarney:
@pcraig3 --- turns out 503 seems to be the status code to send 

Service Unavailable which should get crawlers to return later.

pcraig3:
> Service Unavailable which should get crawlers to return later.

Okay, so return a 503 always as the status code and then don't mess with the "noindex" command at all?

timarney:
> > Service Unavailable which should get crawlers to return later.
> 
> Okay, so return a 503 always as the status code and then don't mess with the "noindex" header at all?

Correct

PR #436, opened 2022-01-06
Cache invalidation on updated content

dsamojlenko:
# Summary | Résumé

Adds a Cache invalidation plugin and related config to force refresh on updated content. 

When a page is updated, this plugin will send cache invalidation requests to CloudFront for the page itself, and any relevant index or category pages.

**Waiting for #442 to get merged so that the IAM user and policy are deployed to Production, then the credentials should be added to github secrets before merging this.** 

There is one minor unrelated change included, see note: https://github.com/cds-snc/gc-articles/pull/436/files#r779654926

PR #439, opened 2022-01-06
Add cache buster IAM user and policy

dsamojlenko:
# Summary | Résumé

Adds the necessary infra changes (IAM user and policy) to support the cache invalidation plugin to be added in #436.


patheard (approved):
Looks good - added a non-blocking comment about `aws_iam_policy_document`

patheard (approved):

PR #452, opened 2022-01-11
Parameterize performance insights

dsamojlenko:
# Summary | Résumé

Performance Insights needs to be enabled in Production, but disabled in Staging.


patheard (approved):
Looks good!  I confirmed that the failed Snyk test is something failing on their end since there are no issues with that file:
image

PR #457, opened 2022-01-12
Activate WPML on new sites

dsamojlenko:
# Summary | Résumé

As part of our efforts to more fully automate onboarding new sites/collections, this PR will automatically configure WPML on new site creation.

It does this by calling directly into the endpoint classes WPML uses in its own setup process.

## Refactor

Included in this PR is a bit of refactoring - moved the contents of `CDS\Modules\Cleanup\Wpml` into this module to keep all the Wpml stuff together.

## Environment variable

Requires a new environment variable:

```
WPML_SITE_KEY=
```

## To test

- Make sure you have the new Environment variable configured
- Create a new site
- Visit that site, notice that WPML is already configured

## Bonus unrelated feature

Added some environment variables to help settle down console logs when working locally in the docker-compose environment. You can now optionally add the following environment variables to turn off logging for certain services:

```
LOCAL_APACHE_LOG_DRIVER=none
LOCAL_PHPMYADMIN_LOG_DRIVER=none
LOCAL_DB_LOG_DRIVER=none
```

**Note this seems to be a bit buggy** - Apache randomly hangs from time to time when logging is turned off, and the only way to recover is to stop/restart the services. So for now, the default is for logging to remain on, but you can try adding these env vars as YMMV.


timarney:
Tested locally --- 

I end up on a "white screen"

Seeing this in the logs
```
error Table 'wordpress.wp_9_icl_string_packages' doesn't exist for query ( \n\t\t\tSELECT\n\t\t\ttranslation_status.rid AS id, 'post'
```

Note the site still gets created and WPML is active

timarney (approved):

PR #472, opened 2022-01-19
Add IDs for translated pages to REST response

pcraig3:
# Summary | Résumé

Returning pages from the API now includes 2 new keys:
- `id_en`
- `id_fr`

They return an integer ID, or 'null' if no translation has been provided.

The most helpful documentation for this was the WordPress docs themselves,
and the translation IDs come from a WPML filter.

Sources:
- https://developer.wordpress.org/rest-api/extending-the-rest-api/modifying-responses/#adding-custom-fields-to-api-responses
- https://wpml.org/wpml-hook/wpml_object_id/


timarney:
Nice tested locally 

Just for clarity this doesn't work when using a slug correct 
```
http://localhost/test2/wp-json/wp/v2/pages/?slug=page-1 (doesn't work)
```

vs 
```
http://localhost/test2/wp-json/wp/v2/pages/22 (works)
```

timarney (approved):

PR #473, opened 2022-01-20
Update code to handle preview links from WP list view

timarney:


pcraig3 (approved):
Works for me! Once drafts are working, it should be good.

PR #477, opened 2022-01-21
Add WAF rule exception for login path

dsamojlenko:
# Summary | Résumé

Adds a WAF rule exception for the login path.

Previously was erroring on password reset submissions.


pcraig3 (approved):
🤞

PR #478, opened 2022-01-24
WIP - add filter to remove wrapping div

timarney:
Screen Shot 2022-01-24 at 10 10 40 AM


pcraig3 (approved):
Seems like we are good to go here 👍

PR #480, opened 2022-01-24
Add theme json

timarney:
Adds custom colours for the CDS redirector theme.

To test insert a "p" tag or another block --- open the inspector to change the colour. Note there should be 2 theme colours to select from.


pcraig3 (approved):
Let's merge! 👍

PR #498, opened 2022-01-26
Feature/reading index settings

timarney:
Adds the HomePage + Site Indexing option to the site settings page.

Screen Shot 2022-01-26 at 12 18 45 PM

### DIFF
--- a/cypress/integration/site.settings.spec.js
+++ b/cypress/integration/site.settings.spec.js
@@ -10,7 +10,7 @@ describe('Site Settings', () => {
   it('Displays settings screen', () => {
     cy.visit('/wp-admin/options-general.php?page=collection-settings');
 
-    cy.get('h1').should('have.text', 'Collection Settings');
+    cy.get('h1').should('have.text', 'Site Settings');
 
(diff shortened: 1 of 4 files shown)


timarney:
Note: 

Also fixes

https://github.com/cds-snc/gc-articles-issues/issues/217

pcraig3 (changes requested):
Looks to be working fine, I have one change request and one question for you.

## Change request: top admin menu link 

I think we should leave in the menu option in the admin that lets people visit the site and come back to the admin. I use it constantly for getting between them. The current link we have at the top only points to the admin, so it is a one-way link.

Screen Shot 2022-01-26 at 15 02 18

## Question: Site name
Should we let them customize the site name as well? It's their site. (Obviously, we don't want them to use bad words or anything, but it seems low risk.) 

Any thoughts, @willeybryan?

Screen Shot 2022-01-26 at 14 58 18

pcraig3 (approved):
The top menu for admins is not related to this PR, and then Bryan's comments we can follow up with in another PR.

PR #500, opened 2022-01-26
Handle when the Theme Option for redirect_url has not been set

dsamojlenko:
# Summary | Résumé

When the option was not set for the redirect_url this bit of code was erroring on the console. This just adds a conditional to prevent the exception. 

Obviously you end up with a broken redirect if you don't fill in that field, so the redirect still doesn't work in the situation, but it's not throwing code exceptions anymore.

Also added some return types and changed the `if/endif`s to `if {}` for consistency

### DIFF
--- a/wordpress/wp-content/themes/cds-redirector/functions.php
+++ b/wordpress/wp-content/themes/cds-redirector/functions.php
@@ -214,12 +214,12 @@ if (!class_exists('Redirector')) {
 Redirector::register();
 
 // Helper function to use in your theme to return a theme option value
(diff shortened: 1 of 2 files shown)


timarney:
@dsamojlenko that works --- wondering if we should do this at the end 

If we know there this isn't ajax + no redirect has been set thinking we should flag it to the user --- could even link back to the setting.

```
if(!$redirectHost){
    die("redirect host not set"); // maybe link to setting
}
```

timarney (approved):

PR #510, opened 2022-01-31
Augment the GC Articles top menu link

pcraig3:
# Summary 

This is the first of my PRs about the admin header, this one is much easier to review + approve.

Previously, GC Admins didn't have an easy way to navigate to the public site, so I augmented the "GC Articles" header menu item. The "GC Articles" link now has the name of the site included, and toggles between the admin interface and the public interface.

It also has a drop-down with options to view the site or the dashboard, so all our bases are covered.

Super admins can also see an "Edit site" link (not pictured), which takes you to the network site settings. After adding this, it made sense to remove the other 'site name' top menu link entirely.

## Screenshots

| dashboard view | viewing the site |
|----------------|------------------|
|      Screen Shot 2022-01-28 at 13 34 21      |      Screen Shot 2022-01-28 at 13 34 27       |

PR #513, opened 2022-02-02
Add language switcher to Admin

pcraig3:
# Summary | Résumé

This PR includes a language-switcher for the admin interface it changes the header toggle between both sites.

### Header toggle

See here for a description: https://github.com/cds-snc/gc-articles/pull/510

### Language switcher 

Built a little change-language form and got it working in the WP Admin Bar. Unlike the language switcher on the public facing site, which toggles between english and french pages, this one changes the backend interface between English and French. 

Note that the languages themselves are hardcoded.

I used the [Simple Admin Change Language](https://en-ca.wordpress.org/plugins/simple-admin-language-change/) for inspiration, although I ended up going a significantly different direction. The two major differences in my approach were:
1. Hardcode the two languages. The plugin loops through all 'installed' languages and displays them in a drop-down
2. Use a form to do a server-side language call. The plugin uses an Ajax call, requiring an extra script to be enqueued.

## Screenshot

Screen Shot 2022-01-31 at 10 54 50

## gif 

Form button looks the same as all the others, whether clicking it or tabbing over it.

Screen Recording 2022-01-31 at 10 54 25

### DIFF
--- a/wordpress/wp-content/plugins/cds-base/classes/Modules/Cleanup/AdminBar.php
+++ b/wordpress/wp-content/plugins/cds-base/classes/Modules/Cleanup/AdminBar.php
@@ -13,6 +13,9 @@ class AdminBar
 
         add_action('admin_bar_menu', [$this, 'addCollections'], 21);
         add_action('admin_bar_menu', [$this, 'addAdminToggle'], 21);
+
+        add_action('admin_bar_menu', [$this, 'addLanguageSwitcher'], 21);
+        add_action('admin_post_cds_change_lang', [$this, 'handleLanguageSwitcherResponse']);
     }
 
     public function addAdminToggle($wp_admin_bar): void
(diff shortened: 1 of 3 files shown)


wordpress/wp-content/plugins/cds-base/classes/Modules/Cleanup/AdminBar.php
```diff
@@ -12,6 +12,113 @@ public function __construct()
...
+            )
+        );
+    }
+
+    public function addLanguageSwitcher($wp_admin_bar): void
+    {
+        $user_locale = get_user_locale();
+        $translation_locale = str_contains($user_locale, "en") ? "fr_FR" : "en_US";
+        // Not using i18n function for these because we don't want to translate them
+        $translation_locale_name = str_contains($user_locale, "en") ? "Français" : "English";
```
pcraig3, line 74: The language strings are hardcoded here, so probably it's better to move them out somewhere.

timarney (approved):

PR #518, opened 2022-02-03
Cleanup static methods and use Class instance

dsamojlenko:
# Summary | Résumé

Small refactor to bring this class in line with our "singleton-ish" pattern that we use elsewhere.

### DIFF
--- a/wordpress/wp-content/plugins/cds-base/classes/Modules/Wpml/Wpml.php
+++ b/wordpress/wp-content/plugins/cds-base/classes/Modules/Wpml/Wpml.php
@@ -99,7 +99,7 @@ class Wpml
         ]);
     }
 
-    public function wpmlTranslatePost()
+    protected function wpmlTranslatePost()
     {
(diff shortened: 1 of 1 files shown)


wordpress/wp-content/plugins/cds-base/classes/Modules/Wpml/Wpml.php
```diff
@@ -11,97 +11,96 @@ public static function register()
...
 
-        self::addTranslatedIDsToPages();
+        $instance = new self();
+
+        add_action('rest_api_init', [$instance, 'addTranslatedIDsToPages']);
     }
 
-    private static function getTranslatedPost($post_id, $lang)
+    public function getTranslatedPost($post_id, $lang)
     {
```
pcraig3, line 20: This one can remain private since it's only called in another method, but it's not really that important.

pcraig3 (approved):

PR #522, opened 2022-02-03
Commit admin and plugin locale files

dsamojlenko:
# Summary | Résumé

Admin and Plugin language files need to be committed because they can't be downloaded post-deploy.


pcraig3 (approved):
Seems like we gotta bite the bullet on this one.

PR #526, opened 2022-02-07
Add filter domain to handle profile updates

timarney:
ref: https://github.com/cds-snc/gc-articles-issues/issues/208

### DIFF
--- a/wordpress/wp-content/plugins/cds-base/classes/Modules/Users/EmailDomains.php
+++ b/wordpress/wp-content/plugins/cds-base/classes/Modules/Users/EmailDomains.php
@@ -27,85 +27,25 @@ class EmailDomains
 
         $pageName = basename($_SERVER['PHP_SELF']);
 
-        if ( 'profile.php' === $pageName){
-           add_filter('is_email', [$this, "isEmail"], 10, 3);
+        if ('profile.php' === $pageName){
+           add_filter('is_email', [$this, "isEmailFilter"], 10, 3);
         }
-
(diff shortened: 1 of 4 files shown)


pcraig3 (changes requested):
The code looks good, but I found two issues when using this:

1. Can't log in using a username anymore.

In practice, this is really about us as admin users. If I type "admin" into the login box, then it looks like the "filterDomain" function gets a boolean value and errors out. (The first sentence there is me printing the value out.)

Screen Shot 2022-02-07 at 10 05 08

2. Editing someone's profile, we get kind of a generic error message.

Screen Shot 2022-02-07 at 10 13 36

It would be good if we could use something more descriptive here.

timarney:
> Editing someone's profile, we get kind of a generic error message.

@pcraig3 --- not seeing a clear way to get the messages updated re: Editing someone's profile, we get kind of a generic error message.

pcraig3:
Now logging in as an admin works, but I am able to edit someone's email address and give them, for example, a "gmail" address.

The name doesn't change at the top of the profile because the `username` is still set to the original email string, but down below, you can see the email has changed.

Screen Shot 2022-02-07 at 14 56 23

timarney:
https://github.com/cds-snc/gc-articles/pull/526#issuecomment-1031860297

Updated to add filter to the user-edit page.

The filter is now applied to profile pages and the user-edit page

pcraig3 (approved):
Looks good! Error message is still kind of generic, but it all works now.

PR #528, opened 2022-02-07
Add "lang" key to pages returned from the API

pcraig3:
# Summary | Résumé

This PR adds a `lang: en|fr` key/value pair to the API response for a page. If somehow no language has been set (eg, WPML is not set up), it will return `lang: null`. 

Usually we know the language when we are calling a page, but not necessarily. Some pages have the same slug in english and french, and if we just call the API without a language parameter, we won't know what language is actually coming back. 

After this PR, each page will know what language its in.

### DIFF
--- a/wordpress/wp-content/plugins/cds-base/classes/Modules/Blocks/src/styles/public.css
+++ b/wordpress/wp-content/plugins/cds-base/classes/Modules/Blocks/src/styles/public.css
@@ -3,8 +3,9 @@ main {
 }
 
 /* Headings */
-.homepage h2 {
-  font-size: 1.2em;
+.homepage h2 ,
+.blog h2 {
+    font-size: 1.2em;
 }
 
 /* WET Menu */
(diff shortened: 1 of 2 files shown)


wordpress/wp-content/plugins/cds-base/classes/Modules/Blocks/src/styles/public.css
```diff
@@ -3,8 +3,9 @@ main {
 }
 
 /* Headings */
-.homepage h2 {
-  font-size: 1.2em;
+.homepage h2 ,
+.blog h2 {
```
timarney, line 7: Just for reference what's this doing in this PR?

timarney (approved):

wordpress/wp-content/plugins/cds-base/classes/Modules/Blocks/src/styles/public.css
pcraig3: The "index" page for me has the big headers again, which breaks the WET styling.

PR #532, opened 2022-02-08
Hide Yubikeys option in personal profile

pcraig3:
# Summary | Résumé

This PR removes the "security keys" part of the two-factor plugin, and a bit of text about Yoast SEO which precedes it.

The Security Keys don't work in Chrome/Edge since Feb 2022, although there is a PR open on the repo to enable this. For now though, we should remove them as an option.

⚠️ Note that existing security keys can't be removed once this is released, let's remove them beforehand ⚠️ 

## Screenshots

| before | to remove | after |
|--------|-----------|-------|
|  Screen Shot 2022-02-08 at 09 26 42   |    Screen Shot 2022-02-08 at 09 26 42 copy    |   Screen Shot 2022-02-08 at 09 27 03   |


timarney:
I think you'll be able to remove the option here under configureProviders.

Screen Shot 2022-02-08 at 9 58 50 AM

PR #535, opened 2022-02-08
Configure staging deployments from manifest

dsamojlenko:
# Summary | Résumé

Updates the Staging deployments to use the new environments manifest. The `environments.yml` file tracks container versions deployed to staging and production environments. Example:

```
production:
  apache: 1.0.19
  wordpress: 2.14.4
staging:
  apache: 1.0.19
  wordpress: 2.15.1
```

## Infrastructure updates
For staging, nothing changes- infrastructure changes (including container deployments) are deployed automatically on merge to main using the same path-based triggers as before.

## Container deployments
Previously, the ECS module's `terragrunt.hcl` would be updated with the new container version to trigger a container deployment, which would happen on merge. 

Now, the container version does not get updated directly in `terragrunt.hcl`, as it is [now referenced from `environments.yml`](https://github.com/cds-snc/gc-articles/pull/535/files#diff-99666f57e53c9c5d39877ab690737fadaa88271195ec64fcd77300455b8d386fR76).

For container deployments, the flow remains the same - we use our same scripts, but the environments.yml file gets updated instead of the terragrunt.hcl file. 

The container update still gets deployed on merge to main like any infrastructure update, because we've added environments.yml to the triggers.

To prevent extra workflow runs when a production version is updated, the new environments-manifest workflow (#543) has a property that lets us know if this is a Prod deployment. If environments.yml is updated, but it's only a production version bump, don't run the staging workflows. (The production version of this will *only* run when that flag is true)

## Deployment script updates

The deployment scripts that we use locally to create the version bump and release PRs work the same, but have been modified to update the `environments.yml` file instead of `terragrunt.hcl`


infrastructure/terragrunt/env/prod/ecs/terragrunt.hcl
```diff
@@ -69,11 +73,11 @@ inputs = {
 
   wordpress_repository_arn = dependency.ecr.outputs.wordpress_repository_arn
   wordpress_image          = dependency.ecr.outputs.wordpress_repository_url
-  wordpress_image_tag      = "v2.14.4"
+  wordpress_image_tag      = "v${local.environments.production.wordpress}"
```
timarney, line 76: 👍

package.json
```diff
@@ -6,6 +6,7 @@
     "deploy-plugin": "./deploy/plugin",
     "update-version": "node ./scripts/prepare-release.js --version_num",
     "tag-release": "node ./scripts/prepare-release.js --tag",
+    "just-testing": "node ./scripts/prepare-release.js --test",
```
timarney, line 9: This will need to be cleaned up.

scripts/util/update-files.js
```diff
@@ -58,13 +60,25 @@ const updatePackageJsonVersion = async (version) => {
...
 }
 
+export const updateEnvironmentManifest = async (version, environment = 'staging') => {
+  const environmentsFile = './infrastructure/environments.yml';
+  const current = fs.readFileSync(environmentsFile, 'utf8')
+  let manifest = YAML.parse(current)
+  manifest[environment].wordpress = version;
+
+  let output = YAML.stringify(manifest);
+  fs.writeFileSync(environmentsFile, output);
```
timarney, line 70: Thinking we don't need the output var here --- just pass YAML.stringify(manifest) into the teFileSync(

scripts/prepare-release.js
```diff
@@ -64,9 +68,9 @@ const inputReleaseTag = async () => {
             await gitCheckMain();
             await gitCheckClean();
             await gitPullLatestFromMain();
-            const { tag, notes } = await inputReleaseTag();
+            const { version, tag, notes } = await inputReleaseTag();
             await gitCreateReleaseBranch(tag);
-            await updateTerragruntHcl(tag);
+            await updateEnvironmentManifest(version, 'staging');
```
timarney, line 73: Maybe your looking to do this later --- should we be pulling in "staging" as a cli arg here?

.github/workflows/environments-manifest.yml
```diff
@@ -0,0 +1,53 @@
...
+        run: |
+          PROD_WORDPRESS=$(yq '.production.wordpress' < ./infrastructure/environments.yml)
+          STAG_WORDPRESS=$(yq '.production.apache' < ./infrastructure/environments.yml)
+          echo "$(cat ./infrastructure/environments.yml)"
+          echo "::set-output name=PROD_WORDPRESS::${PROD_WORDPRESS}"
+          echo "::set-output name=STAG_WORDPRESS::${STAG_WORDPRESS}"
+      - name: Previous versions
+        id: previous
+        run: |
+          PREV_SHA="$(git show -2 --pretty=format:"%h" --no-patch infrastructure/environments.yml | tail -n 1)"
```
patheard, line 47: A risk this has for PRs to Staging is that this only gets the previous commit in the PR branch, so if `infrastructure/environments.yml` is updated multiple times in a single PR, you'll get inconsistent behaviour.

A way around this would be to compare the PR branch version of `infrastructure/environments.yml` to whatever is in `main`.

.github/workflows/terragrunt-deploy-staging-container.yml
```diff
@@ -0,0 +1,85 @@
...
+      - uses: cds-snc/paths-filter@v2.10.3
+        id: filter
+        with:
+          filters: |
+            ecs:
+              - 'infrastructure/terragrunt/aws/ecs/**'
+              - 'infrastructure/terragrunt/env/staging/ecs/**'
+
+      # Load-balancer & database dependency
+      - name: Terragrunt apply ecs
```
patheard, line 82: I think you'll end up with a double `terragrunt apply` of the ECS module here because it still exists in the current `terragrunt-apply-staging.yml` workflow:
https://github.com/cds-snc/gc-articles/blob/main/.github/workflows/terragrunt-apply-staging.yml#L135-L139

dsamojlenko: I don't think so - I tested the triggers here (of course possible I messed up the test but I think this is right) - they don't notice when environments.yml is updated. So unless there's a change in a tf file in the ecs module, this (and the plan workflow) shouldn't be triggered. I will do a bit more testing around this to confirm.

scripts/util/update-files.js
dsamojlenko: true

scripts/util/tag-files.js
```diff
@@ -2,17 +2,22 @@ import shell from 'shelljs';
...
     }
     catch (err) {
-        console.log(err)
+        console.log(err);
     }
 }
 
+export const getVersionTag = async (fileName = "VERSION") => {
+    const version = await getVersion(fileName);
+    return { number: version, tag:`v${version}` };
```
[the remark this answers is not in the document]
dsamojlenko: took a couple iterations to get there but I like it too

package.json
dsamojlenko: oops yup

PR #537, opened 2022-02-09
Add confirmation type to user add form

timarney:
Adds option to send a welcome email when setting up a new user.

I have also added a temporary "Welcome email"

Screen Shot 2022-02-09 at 10 46 44 AM

### DIFF
--- a/wordpress/wp-content/plugins/cds-base/classes/Modules/Users/Users.php
+++ b/wordpress/wp-content/plugins/cds-base/classes/Modules/Users/Users.php
@@ -169,6 +169,33 @@ class Users
         }
     }
 
+    // note this is only available to super admins
+    // for use when setting up a new site / user
+    public function sendWelcome($userExists)
+    {
+        if(!$userExists){
+            $userInfo = get_userdata($uId);
+            $unique = get_password_reset_key($userInfo);
+            $uniqueUrl = network_site_url(
+                "wp-login.php?action=rp&key=$unique&login=" . rawurlencode($userInfo->user_login),
+                'login'
+            );
+        }
+
+        // phpcs:disable
+        $subject = __("Your GC Articles site is ready", "cds-snc");
+        $message = __('@todo add welcome information here', "cds-snc") . "\r\n\r\n";
+        $message .= __('To set your GC Articles account password, please visit the following address:', "cds-snc") . "\r\n\r\n";
+        
+        if(!$userExists){
+            // send reset pass url
+            $message .= $uniqueUrl;
+        }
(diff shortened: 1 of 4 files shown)


wordpress/wp-content/plugins/cds-base/classes/Modules/Users/Users.php
```diff
@@ -163,6 +169,33 @@ public function addToBlog($uId, $role)
...
+        // phpcs:disable
+        $subject = __("Your GC Articles site is ready", "cds-snc");
+        $message = __('@todo add welcome information here', "cds-snc") . "\r\n\r\n";
+        $message .= __('To set your GC Articles account password, please visit the following address:', "cds-snc") . "\r\n\r\n";
+        
+        if(!$userExists){
+            // send reset pass url
+            $message .= $uniqueUrl;
+        }
+        // phpcs:enable
```
pcraig3, line 194: If the user exists already, there's no link, eh? Should we link them to the login page for the new site?

pcraig3: Or does the existing message cover it? (eg, maybe we don't need a welcome message unless it's a totally new user?)

wordpress/wp-content/plugins/cds-base/classes/Modules/Users/src/UserForm.tsx
```diff
@@ -187,6 +189,37 @@ export const UserForm = (props) => {
...
+                                <td>
+                                    <label style={{ marginRight: "12px" }}>
+                                        <input
+                                            name="confirmationType"
+                                            type="radio"
+                                            value="welcome"
+                                            checked={confirmationType === "welcome"}
+                                            onChange={(event) => { bindConfirmationType.onChange(event) }}
+                                        />
+                                        {__(" Welcome", "cds-snc")}
```
pcraig3, line 208: One extra space here.

pcraig3 (approved):
All looks good to me. Left a few comments about what happens if the user already exists, but nothing that would stop this from going it.

wordpress/wp-content/plugins/cds-base/classes/Modules/Users/Users.php
timarney: Was thinking about a case where this could be the 2nd or 3rd site the person has access to.  In this case we don't want to them to reset the password.

In which case I should remove this as well.

> $message .= __('To set your GC Articles account password, please visit the following address:', "cds-snc") . "\r\n\r\n";

PR #564, opened 2022-02-16
Environments Manifest: Use prev changed ref when on main

dsamojlenko:
# Summary | Résumé

Updates the environments manifest ref for previous versions. 

#560 failed because once the manifest file has merged to main, we were just comparing the file against itself and the workflow wouldn't run.

So the Previous Versions job has been updated so:
- If on a PR, use the manifest file in main as the ref
- If on main, use the last change to the manifest file as the ref

This works because:
- When on a PR, the manifest file in main will always be the last deployed versions.
- When on main, the last time the manifest file was updated will always be the last deployed versions.

### DIFF
--- a/.github/workflows/environments-manifest.yml
+++ b/.github/workflows/environments-manifest.yml
@@ -56,8 +56,12 @@ jobs:
       - name: Previous versions
         id: previous
         run: |
-          git branch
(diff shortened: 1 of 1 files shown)


timarney (approved):
As discussed this seems to make sense.

PR #586, opened 2022-02-18
Send welcome email via Notify template

timarney:
Updated code to send "Welcome" email via a new Notify template
 
 https://github.com/cds-snc/gc-articles-issues/issues/242


pcraig3 (approved):
Tested and it works for me 👍

PR #588, opened 2022-02-18
Consistent Admin menus for English and French language settings

pcraig3:
# Summary | Résumé

This PR does a few things: Standardizes admin menu items for EN and FR language settings, removes "Salutations" in the top admin bar header, and fixes a whitespace linting error in our admin css file.

Resolves: https://github.com/cds-snc/gc-articles-issues/issues/247

## Standardizes our menu items

When you were logged in as an administrator, the left hand menu items would change as you switched between English and French. It would also remove all the separators from the menu, which, while it doesn't impact functionality, wasn't something we intended. 

New code standardizes the menu items between both languages, and keeps the separators in place. 

(green underlines added so the changes are clearer.)

| before | after |
|--------|-------|
|        |  added "Dashboard", "Les Articles", and "Envoyer en masse" . Also "salutations" is gone (top, right)    |
|   Screen Shot 2022-02-18 at 14 46 08   |  Screen Shot 2022-02-18 at 14 45 29  |

| before | after |
|--------|-------|
|        |  added "Dashboard", and menu items are spaced out more     |
|  Screen Shot 2022-02-18 at 14 46 02    |  Screen Shot 2022-02-18 at 14 45 34    |

## Removes "Salutations"

The wordpress profile link normally says "Howdy", but we removed that. Now that we have a French admin UI, it was saying "Salutations", and so we're removing that too.

### DIFF
--- a/wordpress/wp-content/plugins/cds-base/classes/Modules/Cleanup/Menus.php
+++ b/wordpress/wp-content/plugins/cds-base/classes/Modules/Cleanup/Menus.php
@@ -25,30 +25,45 @@ class Menus
             return;
         }
 
-        global $menu, $submenu;
+        global $menu;
 
         /* add items to keep here */
         $allowed = [
-            __('Pages'),
-            __('Posts'),
-            __('Articles', 'cds-snc'),
-            __('Bulk', 'cds-snc'),
-            __('Users'),
-            __('Settings'),
-            __('Media')
+            // English menu items
+            "Dashboard",
+            "Articles",
+            "Media",
+            "Pages", // same in FR
+            "Menus", // same in FR
+            "Users",
+            "Settings",
+            "Bulk Send",
+
+            // French menu items
+            "Tableau de bord",
+            "Les Articles",
+            "Média",
+            "Utilisateurs",
+            "Réglages",
+            "Envoyer en masse",
         ];
 
         //  __('Settings'), __('Appearance')
         // http://localhost/wp-admin/options-reading.php
         end($menu);
         while (prev($menu)) {
-            $value = explode(' ', $menu[key($menu)][0]);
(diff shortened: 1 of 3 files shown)


wordpress/wp-content/plugins/cds-base/classes/Modules/Cleanup/Menus.php
```diff
@@ -25,30 +25,45 @@ public function removeMenuPages(): void
...
+            "Settings",
+            "Bulk Send",
+
+            // French menu items
+            "Tableau de bord",
+            "Les Articles",
+            "Média",
+            "Utilisateurs",
+            "Réglages",
+            "Envoyer en masse",
```
pcraig3, line 48: Note that I removed the i18n wrapper here because we aren't using these strings in the interface, but to do a logical comparison.

wordpress/wp-content/plugins/cds-base/classes/Modules/Cleanup/Menus.php
```diff
@@ -25,30 +25,45 @@ public function removeMenuPages(): void
...
-            $value = explode(' ', $menu[key($menu)][0]);
-            if (! in_array($value[0] !== null ? $value[0] : '', $allowed)) {
+            $value = $menu[key($menu)] ?? [];
+            $label = $value[0] ?? '';
+            $isSeparator = ($value[4] ?? null)  === "wp-menu-separator";
+
+            // Unset menu items unless they are safelisted or separators
+            if (!in_array($label, $allowed) && !$isSeparator) {
                 unset($menu[key($menu)]);
             }
```
pcraig3, line 62: Cleaned up the logic a bit here, and kept the menu separators.

PR #589, opened 2022-02-21
Auto create fr content - finish site setup

timarney:
Adds code to pull and add FR content for pages

To test --- add 2 translated pages to "template-site"

Adds translated content for:

- home ... slug = home-fr
- maintenance = maintenance-fr

Previous setup

```
Make a site will pull content from called -> template-site
Under that site
make a "home page" with the slug -> home
make a "maintenance page" with the slug -> maintenance

Those sites will be the templates for the new site you create

```


timarney:
ref: https://github.com/cds-snc/gc-articles-issues/issues/224

dsamojlenko (approved):

PR #590, opened 2022-02-21
Remove memsource

timarney:
Closes https://github.com/cds-snc/gc-articles-issues/issues/238

### DIFF
--- a/wordpress/composer.json
+++ b/wordpress/composer.json
@@ -120,12 +120,5 @@
     "psr-4": {
       "CDS\\": "wp-content/plugins/cds-base/classes/"
     }
-  },
-  "config": {
-    "allow-plugins": {
-      "composer/installers": true,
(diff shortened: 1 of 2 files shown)


wordpress/composer.json
```diff
@@ -121,5 +120,12 @@
     "psr-4": {
       "CDS\\": "wp-content/plugins/cds-base/classes/"
     }
+  },
+  "config": {
+    "allow-plugins": {
+      "composer/installers": true,
+      "ffraenz/private-composer-installer": true,
+      "pestphp/pest-plugin": true
+    }
```
dsamojlenko, line 129: what does this do?

timarney: Will remove --- think that was from a bad update

dsamojlenko (approved):

PR #603, opened 2022-03-01
Update Contact form

timarney:
# Summary | Résumé

Updates contact form + removes request a site steps (which will be it's own form)


wordpress/wp-content/plugins/cds-base/classes/Modules/Contact/ContactForm.php
```diff
@@ -41,227 +60,198 @@ public function render($atts, $content = null): string
...
-
-                    <div class="focus-group">
-                        <label class="gc-label required" for="email" id="email-label">
-                            <?php _e('Email', 'cds-snc'); ?>
-                        </label>
-                        <input type="email" class="gc-input-text" id="email" required autocomplete="email"
-                               name="email" value="">
-                    </div>
+        <!--
+            fullname
```
pcraig3, line 69: We can probably get rid of this one too.

wordpress/wp-content/plugins/cds-base/classes/Modules/Contact/ContactForm.php
```diff
@@ -16,6 +16,25 @@ public function register()
...
     }
 
+    public function checkBox($name, $id, $value): void
+    {
+        ?>
+        <div class="gc-input-checkbox">
+            <input
+                    name="<?php echo $name; ?>"
+                    class="gc-input-checkbox__input"
+                    id="<?php echo $id; ?>"
```
pcraig3, line 26: Ooh, this breaks web conventions [according to MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/id):

> id's value must not contain [whitespace](https://developer.mozilla.org/en-US/docs/Glossary/Whitespace) (spaces, tabs etc.). Browsers treat non-conforming IDs that contain whitespace as if the whitespace is part of the ID. In contrast to the [class](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes#attr-class) attribute, which allows space-separated values, elements can only have one single ID value.

Ours are just full sentences.

Screen Shot 2022-03-01 at 16 04 55

We could strip all the whitespace and we would still get unique ids.

wordpress/wp-content/plugins/cds-base/classes/Modules/Contact/ContactForm.php
```diff
@@ -41,227 +60,198 @@ public function render($atts, $content = null): string
...
-                <form id="contact-form" method="POST" action="/wp-json/contact/v1/process">
-                    <?php // echo $heading; ?>
-                    <?php wp_nonce_field(
-                        'contact_form_nonce_action',
-                        'contact',
+                    <p><?php _e('We use this information to improve GC Articles.', 'cds-snc');?></p>
+                    
+                    <div class="focus-group" style="margin-bottom: 1.75rem">
+                    <?php $this->checkBox(
+                        'usage[]',
```
pcraig3, line 158: Is the "[]" bit significant? I guess it lets us know that there could be multiple values?

wordpress/wp-content/plugins/cds-base/classes/Modules/Contact/ContactForm.php
```diff
@@ -41,227 +60,198 @@ public function render($atts, $content = null): string
...
                     </div>
-                    <label data-testid="description" class="gc-label required" id="contact-label" for="message">
-                        <?php echo $label; ?>
+                    <input 
+                        type="text" 
+                        class="gc-input-text" 
+                        id="usage-other" 
+                        name="usage-other" 
+                        value=""
+                    />
```
pcraig3, line 189: These are getting flagged in axe for not having labels. We could create a visually hidden label that says "Other usage" or something.

wordpress/wp-content/plugins/cds-base/classes/Modules/Contact/ContactForm.php
```diff
@@ -41,227 +60,198 @@ public function render($atts, $content = null): string
...
+                    ); ?>
                     </div>
-                </form>
+                    <input 
+                        type="text" 
+                        class="gc-input-text" 
+                        id="target-other" 
+                        name="target-other" 
+                        value=""
+                    />
```
pcraig3, line 233: Same as the other one, this gets flagged for no `<label>`.

wordpress/wp-content/plugins/cds-base/classes/Modules/Contact/ContactForm.php
```diff
@@ -41,227 +60,198 @@ public function render($atts, $content = null): string
...
+                    <?php _e('Your message', 'cds-snc'); ?>
+                </label>
+                <textarea
+                    data-testid="textarea"
+                    class="gc-textarea"
+                    id="message"
+                    required
+                    placeholder=""
+                    name="message"
+                ></textarea>
```
pcraig3, line 246: Both `id`s are "message". I guess let's change one.

pcraig3 (changes requested):
I've added some comments above (or below — not sure how the interface handles it).

I've also noticed a couple of weird things in the email template.

## 1. The "message" content is really far from the message heading

The answers for my target audience, and my intended usage always come before the "message" content for some reason.

Screen Shot 2022-03-01 at 16 01 23


## 2. (Other) subheadings always show up

The (Other) from the "target audience" and the "intended usage" questions always show up in the email template, even if I don't answer those questions.

Screen Shot 2022-03-01 at 16 00 59

pcraig3:
Also, should it translate into french? It doesn't seem like it's working for me locally.

timarney:
> **Also, should it translate into french? It doesn't seem like it's working for me locally.**

We'll need to compile all the updates

wordpress/wp-content/plugins/cds-base/classes/Modules/Contact/ContactForm.php
timarney: This will get us an array in PHP
https://stackoverflow.com/questions/4516847/make-array-from-checkbox-form

timarney:
> 1. The "message" content is really far from the message heading

Updated the Notify template 
Screen Shot 2022-03-02 at 6 56 12 AM

timarney:
FYI - @pcraig3 ... I'm working on the ZenDesk integration for this now in the PR.

wordpress/wp-content/plugins/cds-base/classes/Modules/Contact/ContactForm.php
pcraig3: Neat! I didn't know that.

mohamed-cds (commented):
Dee in internal ops noticed that the tickets from articles are being misrouted and are being assigned to the wrong group (due to zendesk defaults). Can you pass a tag before this PR is merged so i can setup some basic routing `'tags' => ['articles_api']` under `request` should do the trick

pcraig3 (approved):

pcraig3:
Approved! Let's merge it in and keep going.

PR #604, opened 2022-03-01
Add options to side search bar and breadcrumbs

timarney:
# Summary

This PR restructures the settings page, and adds 2 new settings: 1 to hide the breadcrumbs, and 1 to hide the search bar.

## 1. Restructure the settings page

It was getting a bit cluttered, so I started grouping the various options under headings.

| before | after |
|--------|-------|
|   Screen Shot 2022-03-02 at 12 13 59    |   Screen Shot 2022-03-02 at 12 13 27  |

Changes:
- I created 3 sub-sections — we can change them if we want — but there you go
- Added a couple of sentences to explain what maintenance mode is
- Removed the double "settings saved" flash message
- Added the 2 new settings at the bottom
- Moves the "Site settings" and "Notify API Settings" options to the bottom of the Settings panel 

### Hide the search bar

Hide the search bar if not needed. Adds a CSS class to the body which adds `display: none` to the search bar.

### Hide the breadcrumbs

Hide the breadcrumbs if not needed. Adds a CSS class to the body which adds `display: none` to the search bar.

I spent a while looking into the Yoast setting that Enables or Disables the breadcrumbs. The yoast breadcrumb setting is stored in a nested option that you can't update with a normal implementation of the settings API.

You have to call `get_option('wpseo_titles')['breadcrumbs-enable']` to get the Yoast setting. Since I wasn't able to figure it out, I created a new setting `show_breadcrumbs`, which will add a CSS class to the body, same as the search bar setting.

### Screenshots

Desktop

| with search bar and breadcrumbs |  hidden search bar and breadcrumbs  |
|--------|-------|
|  Screen Shot 2022-03-02 at 12 33 38  |  Screen Shot 2022-03-02 at 12 34 24  |

Mobile

| with search bar and breadcrumbs |  hidden search bar and breadcrumbs  |
|--------|-------|
|   Screen Shot 2022-03-02 at 12 34 49   |   Screen Shot 2022-03-02 at 12 34 47  |

### DIFF
--- a/wordpress/wp-content/plugins/cds-base/classes/Modules/Site/SiteSettings.php
+++ b/wordpress/wp-content/plugins/cds-base/classes/Modules/Site/SiteSettings.php
@@ -90,6 +90,11 @@ class SiteSettings
 
         register_setting(
             'site_settings_group', // option_group
+            'show_search',
+        );
+
+        register_setting(
+            'site_settings_group', // option_group
             'blogname',
         );
 
@@ -107,6 +112,17 @@ class SiteSettings
             'collection_settings_section', // section
             [
                 'label_for' => 'collection_mode'
+            ]
+        );
+
+        add_settings_field(
+            'show_search', // id
+            __('Show Search', 'cds-snc'), // title
+            array( $this, 'showSearchCallback'), // callback
+            'collection-settings-admin', // page
+            'collection_settings_section', // section
+            [
+                'label_for' => 'show_search'
             ]
         );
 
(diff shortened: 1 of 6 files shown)


wordpress/wp-content/plugins/cds-base/classes/Modules/Site/SiteSettings.php
```diff
@@ -40,7 +40,6 @@ public function collectionSettingsCreateAdminPage()
 
         <div class="wrap">
             <h1><?php _e('Site Settings', 'cds-snc') ?></h1>
-            <?php settings_errors(); ?>
```
timarney, line 43: Wondering why this was removed.

timarney: Is this why the double messages were showing?

pcraig3: Yeah, removed this and the double errors/confirmation messages went away.

pcraig3 (approved):
Great job, team!

PR #615, opened 2022-03-03
Add GC Writer role

timarney:
I'll look to add some more specific tests i.e. this user can't publish ... but wanted to get this code in first.

Closes https://github.com/cds-snc/gc-articles-issues/issues/182

### DIFF
--- a/wordpress/wp-content/plugins/cds-base/classes/Modules/Users/Users.php
+++ b/wordpress/wp-content/plugins/cds-base/classes/Modules/Users/Users.php
@@ -61,16 +61,16 @@ class Users
         $role_names_arr = [];
 
         $administrator = __(
-            "This role has complete control over the articles collection and can perform all other roles actions",
+            "This role has complete control over the articles site and can perform all other roles actions",
             "cds-snc"
         );
         $gceditor = __(
-            "This role is allows the user to write and publish articles online to the collection.",
+            "This role allows the user to write and publish articles.",
             "cds-snc"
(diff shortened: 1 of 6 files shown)


wordpress/wp-content/plugins/cds-base/classes/Modules/Users/Users.php
```diff
@@ -68,7 +68,12 @@ public function getRoles()
             "This role is allows the user to write and publish articles online to the collection.",
             "cds-snc"
         );
-        $roleDescriptions = ["administrator" => $administrator, "gceditor" => $gceditor];
+
+        $gcwriter = __(
+            "This role is allows the user to write articles online to the collection.",
+            "cds-snc"
+        );
```
pcraig3, line 75: - don't need the "is" in both descriptions
- we should probably change this to not say "collection"?

Also, the main difference between this one and an Editor is that this one can't publish, right? Should that be more clear in the description? ("This role allows the user to draft articles but not to publish them.")

pcraig3 (commented):
Looks good. 

There _is_ still a "Publish" button for me but when you press it, it says "submit for review" instead, so all good.

When I log in again as an admin, I see this "someone else is editing this post" modal, which could be confusing for someone if it happens in the wild. I say we just merge and keep an eye if anyone raises it in the future.

Screen Shot 2022-03-07 at 15 45 50

pcraig3 (approved):

PR #616, opened 2022-03-04
Request a site

pcraig3:
# Summary 

This pull request adds a new "Request a site" form which is super copy + pasted from the Contact form.

It's not my best work, but it should be okay for what we need.

Todo after this is merged:
- Show/hide "other" fields
- Do a 'URL preview' widget
- Translations
- Merge this with Contact form to reduce copy and paste (?)

# Screenshots

| first step | second step |
|--------|-------|
| localhost_request-a-site-please_  | localhost_request-a-site-please (2)     |

| first step (with error) | second step (with error) |
|--------|-------|
| Note that the missing field doesn't have a validation message (this how it works, unfortunately) | error at the top and validation as well |
|    localhost_request-a-site-please    |  localhost_request-a-site-please (1)   |


timarney:
@pcraig3 

We'll need to add the block (build) to the preinstall here (not seeing that in this PR)

https://github.com/cds-snc/gc-articles/pull/598/files#diff-db981b357e2880e26136e7e36b8a8d0fa3ceb83a921efb73f135b1c8d7adf694R43

wordpress/wp-content/plugins/cds-base/classes/Modules/FormRequestSite/Setup.php
```diff
@@ -0,0 +1,122 @@
...
+
+        new RequestSite();
+        new Block();
+    }
+
+    public function enqueue()
+    {
+        wp_enqueue_script('cds-request-js', plugin_dir_url(__FILE__) . '/src/handler.js', ['jquery'], "1.0.0", true);
+
+        // where is this?
```
timarney, line 41: Believe we can remove this --

CDS_VARS is localized in another script ... this creates an object on the Window (window.CDS_VARs)

wordpress/wp-content/plugins/cds-base/classes/Modules/FormRequestSite/src/handler.js
```diff
@@ -0,0 +1,45 @@
...
+            success: function (data) {
+                console.log(data);
+                if (data && data["error"]) {
+                    var errorEl = '<div id="request-error" class="gc-alert gc-alert--error gc-alert--validation" data-testid="alert" tabindex="0" role="alert">';
+                    errorEl += '<div class="gc-alert__body">';
+                    errorEl += '<h2 class="gc-h3">' + data.error_message + '</h2>';
+                    if(data['keys']) {
+                        errorEl += '<ol class="gc-ordered-list">';
+                        data['keys'].forEach(key => {
+                            errorEl += '<li><a href="#' + key + '" class="gc-error-link">Please complete the required field: ' + key +'</li>'
```
timarney, line 26: We'll need to pass these strings from the sever so we can get them into the correct language.

timarney:
@dsamojlenko does the build / install pipeline for this PR work for you?

I'm getting that path issue again

```
hosted-git-info/git-host.js:47
vars.path = vars.path ? vars.path.replace(/^[/] + / g, '') : ''
```

We'll need to add the build assets ...

[Docker](https://github.com/cds-snc/gc-articles/blob/main/wordpress%2Fdocker%2FDockerfile#L50)

timarney:
Outside of the build assets -- I would say we can approve / merge this and carry on with getting some shared code moved out between the forms.

---

- Send Mail
- ZenDesk
- jQuery

wordpress/wp-content/plugins/cds-base/classes/Modules/FormRequestSite/src/handler.js
pcraig3: Okay, I'm just removing the second string here.

Screen Shot 2022-03-08 at 09 56 08
Screen Shot 2022-03-08 at 09 56 37

PR #617, opened 2022-03-07
Hide desc and xfn text fields for menu items

timarney:
- Updates `admin.css` to hide `description` and `xfn` text fields by default.  
- Adds `admin css` to the Redirector theme to turn `description` back on for that theme.

Closes https://github.com/cds-snc/gc-articles-issues/issues/239


pcraig3 (approved):
Looks good! 👍

PR #618, opened 2022-03-07
Updates Dashboard - pull content from REST API

timarney:
Updates the Dashboard to pull content from the "base website" --- I was having CURL issues pointing to my local host so this is set to the `prod` website --- will be curious if using site_url(1); is broken for others as well or just my box.


pcraig3 (approved):
Works! Let's merge it.

PR #620, opened 2022-03-08
Messenger.php sends emails and creates tickets

pcraig3:
# Summary | Résumé

Added a new class to share code for sending messages between our request form and our contact form. It also adds a "CC" option to the Request a site form, duplicating the contact form behaviour.

Needs doing:
- [x] generalize sendMail method
- [x] add tests for sendMail
- [x] generalize the method to create Zendesk tickets
- [x] add tests for the Zendesk thing
- [x] CC on the request a site form

Notes:

- Mocking the NotifyClient wasn't possible when the `new` object was being created directly in the `sendEmail` function. [The Mockery docs](https://docs.mockery.io/en/latest/cookbook/mocking_class_within_class.html) recommend creating a second function that returns the new object, which is why it's like that.
- Should I make these static calls? Or maybe we want to pass in the template ID later or something
- I put the Messenger class in a directory called `/Forms`, assuming that the others will make it in there eventually. I can rename this if we want to take more of a 'wait and see' approach

### DIFF
--- a/wordpress/wp-content/plugins/cds-base/classes/Modules/Forms/tests/MessengerTest.php
+++ b/wordpress/wp-content/plugins/cds-base/classes/Modules/Forms/tests/MessengerTest.php
@@ -1,34 +1,62 @@
 <?php
 
 use CDS\Modules\Forms\Messenger;
+use GuzzleHttp\Exception\ClientException;
 
 
 test('asserts sendMail returns success message for a successful call', function () {
     /* Mocking a class within a class: https://docs.mockery.io/en/latest/cookbook/mocking_class_within_class.html */
     $clientMock = mock('NotifyClientMock')->expect(
-            sendMail: fn () => true,
+        sendMail: fn () => true,
     );
 
     $messenger = mock(Messenger::class)->makePartial();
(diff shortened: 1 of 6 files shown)


wordpress/wp-content/plugins/cds-base/classes/Modules/Forms/tests/MessengerTest.php
```diff
@@ -0,0 +1,63 @@
...
+
+    expect($messenger->createTicket('Goal: Request a site', 'Paul Craig', 'paul@paul.ca', 'message'))->toMatchArray(
+        ['success' => 'Success']
+    );
+});
+
+test('asserts createTicket returns Zendesk server error generic exception', function () {
+    $clientMock = mock('GuzzleClientMock');
+    $clientMock->shouldReceive('request')->andThrow(new \Exception('Maybe we do need API auth after all'));
+    // $clientMock->shouldReceive('request')->andThrow(new ClientException('Sorry, it is not yet March 19th', $requestMock));
```
timarney, line 53: Let's remove this.

timarney (approved):
Looks great --- one super minor comment but approving this.

pcraig3:
Going to merge, will re-compile strings in the AM.

PR #621, opened 2022-03-09
Add jjQuery accessible toggle for optional fields

pcraig3:
# Summary | Résumé

This PR adds an accessible jQuery toggle for the optional fields on the first page of the Request a site form and on the Contact form.

The JS is duplicated, but that's actually a good thing because we can get rid of one copy in the (near) future.

## fun gif

Screen Recording 2022-03-10 at 13 20 10

## Description

I wasn't able to check my past work on this, but I am pretty confident. The idea is that the controlling element should have 2 aria attributes and the element being controlled should have 1.  There is an example on here that mostly confirms my approach: https://van11y.net/accessible-hide-show/

```html
<!-- expanded -->
<button aria-controls="toggle-area-1" aria-expanded="true">Toggle the stuff</button>
<div id="toggle-area-1" aria-hidden="false">
  <p>This stuff will be toggled 1</p>
</div>

<!-- hidden -->
<button aria-controls="toggle-area-2" aria-expanded="false">Toggle the stuff</button>
<div id="toggle-area-2" aria-hidden="true">
  <p>This stuff will be toggled 2</p>
</div>
```

The controlling element has an "aria-controls" attribute, with the id of the element that gets hidden. It also has an "aria-expanded" boolean attribute that is "true" when visible and "false" when hidden.

The element being toggled has an "id", obviously, and "aria-hidden" that says whether it's visible or not.

I'm not 100% sure that we need aria-hidden, as the example I linked to above doesn't use it. But it doesn't hurt anything.

### DIFF
--- a/wordpress/wp-content/plugins/cds-base/classes/Modules/Blocks/src/styles/public.css
+++ b/wordpress/wp-content/plugins/cds-base/classes/Modules/Blocks/src/styles/public.css
@@ -82,6 +82,6 @@ header .gcweb-menu {
 }
 
 /* Aria toggle */
-form *[aria-hidden="true"] {
+form .displayNone {
   display: none;
 }
--- a/wordpress/wp-content/plugins/cds-base/classes/Modules/Blocks/src/styles/public.css
+++ b/wordpress/wp-content/plugins/cds-base/classes/Modules/Blocks/src/styles/public.css
@@ -80,3 +80,8 @@ header .gcweb-menu {
 .hide-breadcrumbs #wb-bc {
   display: none;
 }
+
+/* Aria toggle */
+form *[aria-hidden="true"] {
+  display: none;
+}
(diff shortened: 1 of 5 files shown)


timarney:
Found an issue where the other usage text field doesn't get announced when you tab it it --- after selecting + deselecting once.

https://www.youtube.com/watch?v=UhALEoRJJ_w

We can merge and file a separate bug for that if you like.

timarney:
Also filed 2 other issues while testing this 

https://github.com/cds-snc/gc-articles-issues/issues/274 
https://github.com/cds-snc/gc-articles-issues/issues/275

pcraig3:
Latest commit resolves https://github.com/cds-snc/gc-articles-issues/issues/274

And it should get voiceover back announcing what we want it to announce. (I tested locally and it does.)

timarney (approved):

PR #624, opened 2022-03-11
Feature/service ids

timarney:
Update List Manager to use Service IDs for pulling and displaying data.

### DIFF
--- a/wordpress/wp-content/plugins/cds-base/classes/Modules/ListManager/app/public/index.html
+++ b/wordpress/wp-content/plugins/cds-base/classes/Modules/ListManager/app/public/index.html
@@ -8,6 +8,6 @@
   </head>
   <body>
     <noscript>You need to enable JavaScript to run this app.</noscript>
-    <div id="list-manager-app"></div>
+    <div id="list-manager-app" data-ids='[{"name":"List One","service_id":"a7902fc7-37f0-419c-84c8-3ab499ee24c8"},{"name":"Service 222","service_id":"9e7cf7e0-d40d-4e43-a3c7-ea84344f4291"}]'></div>
   </body>
 </html>
--- a/wordpress/wp-content/plugins/cds-base/classes/Modules/ListManager/app/src/components/ListForm.tsx
+++ b/wordpress/wp-content/plugins/cds-base/classes/Modules/ListManager/app/src/components/ListForm.tsx
@@ -32,104 +32,100 @@ export const ListForm = ({ handler, formData = {}, serverErrors = [] }: { handle
     return (
         <form onSubmit={handleSubmit(handler)}>
             <table id="form-table" className="form-table">
-                <thead>
-                    <tr>
-                        <th><label className="required" htmlFor="name"><Asterisk />{__("Name", "cds-snc")}</label></th>
-                        <td>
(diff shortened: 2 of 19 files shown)


wordpress/wp-content/plugins/cds-base/classes/Modules/ListManager/app/public/index.html
```diff
@@ -8,6 +8,6 @@
   </head>
   <body>
     <noscript>You need to enable JavaScript to run this app.</noscript>
-    <div id="list-manager-app"></div>
+    <div id="list-manager-app" data-ids='[{"name":"List One","service_id":"a7902fc7-37f0-419c-84c8-3ab499ee24c8"},{"name":"Service 222","service_id":"9e7cf7e0-d40d-4e43-a3c7-ea84344f4291"}]'></div>
```
pcraig3, line 11: Was this left in by accident?

wordpress/wp-content/plugins/cds-base/classes/Modules/ListManager/app/src/components/ListForm.tsx
```diff
@@ -32,100 +32,104 @@ export const ListForm = ({ handler, formData = {}, serverErrors = [] }: { handle
...
-                    <th></th>
-                    <td><input className="button button-primary" type="submit" /></td>
-                </tr>
+                <thead>
+                    <tr>
+                        <th><label className="required" htmlFor="name"><Asterisk />{__("Name", "cds-snc")}</label></th>
+                        <td>
+                            <div className={errors.name ? "error-wrapper" : ""}>
+                                {errors.name && <span className="validation-error">{errors.name?.message || __("Name is required", "cds-snc")}</span>}
+                                <input style={textWidth} type="text" {...register("name", { required: true })} />
```
pcraig3, line 41: These inputs need an `id` as well, otherwise the form label isn't associated with them.

wordpress/wp-content/plugins/cds-base/classes/Modules/ListManager/app/src/components/ListViewTable.tsx
```diff
@@ -177,13 +176,13 @@ export const ListViewTable = () => {
                         Header: 'Upload',
                         accessor: 'active',
                         Cell: ({ row }: { row: any }) => {
-                            return <UploadListLink id={`${row?.values?.id}`} />
+                            return <UploadListLink serviceId={serviceId} listId={`${row?.values?.id}`} />
```
pcraig3, line 179: These are returning "same text for different links" error. Easiest fix is to do "Upload list (to Paul's awesome subscribers)".

pcraig3 (approved):
It all works pretty well for me, so going to pre-approve. 

I made a few comments above, but nothing too serious. 

I am also still seeing the error where my service disappears when I rename it, not sure if you wanted to do that as a follow-up.

wordpress/wp-content/plugins/cds-base/classes/Modules/ListManager/app/public/index.html
timarney: That's hardcoded for the standalone version --- you can spin up React on localhost:3000 outside of WP.

PR #626, opened 2022-03-14
Move Subscribe form into the "Forms" folder + share common functions in "Utils.php"

pcraig3:
# Summary | Résumé

This PR moves the "Subscribe" form into the `Forms` folder, alongside our other forms. It also creates a new `Utils.php` class that mostly contains shared functions to build form fields.

The only visible change is that I removed the list element from the "description" under the "request a site" email address field. Every other field that needs a description just uses a string, so I changed the Email address description to be the same as the others. 

## Moving the Subscribe form 

Didn't change too much about the internal logic to get it to work, although I did change some things.

- I changed some of the form markup so that it's closer to what we use in the contact form
- I also deleted the separate "handler.js", since the other one handles everything already

## Utils.php

The new functions in Utils are:
- `textField`
- `radioField`
- `checkboxField`
- `submitButton`
- `isNonceErrorMessage` → returns error message if the nonce is bad, otherwise, returns empty string

All the other methods should be self-explanatory.

Also added some simple tests for all the form field creation methods.


wordpress/wp-content/plugins/cds-base/classes/Modules/Forms/Subscribe/Setup.php
```diff
@@ -106,27 +94,43 @@ protected function subscribe(string $email, string $listId): array
...
+                'error' =>  true,
+                'error_message' => __('Unknown subscription', 'cds-snc'),
+                'keys' => ['list_id']
+            ];
         }
 
         $email = sanitize_email($_POST["email"]);
         $listId = sanitize_text_field($_POST['list_id']);
 
+        // @TODO mail people when they subscribe
```
timarney, line 132: This should already happen via ListManager.

pcraig3: Okay, good to know.

pcraig3:
This PR, if merged, closes (6) and (7) in this issue: https://github.com/cds-snc/gc-articles-issues/issues/279

wordpress/wp-content/plugins/cds-base/classes/Modules/Forms/Contact/ContactForm.php
```diff
@@ -162,27 +86,27 @@ class="gc-input-text"
                     </div>
 
                     <div class="focus-group">
-                    <?php $this->checkBox(
+                    <?php echo Utils::checkboxField(
```
timarney, line 89: 1) Wondering if we can remove all these opening and closing PHP tags now that this is cleaned up.
2) What do you thinking about echo 'ing but default and having an optional param to return vs echo (might help the readability)

timarney (approved):

pcraig3:
Gonna ship some improvements in a follow-up PR

PR #628, opened 2022-03-15
Save list values from api

timarney:
## PR

ref: https://github.com/cds-snc/gc-articles-issues/issues/281

This PR updates the code to "sync" / save the values when an API call is made to get current list data.

Once the list data is saved visiting `List Manager Settings` should populate the List Data with the saved information from the API call.

Once this PR is merged -- we can make a further update to use the the Notify service from here `options-general.php?page=notify-settings` as the sending service for lists.  Given the decision was made to use a single sending service we'll be able to remove the "List Manager Settings" settings after that - we can remove the sending service dropdown from bulk send and just used the defined service.

Screen Shot 2022-03-15 at 2 33 21 PM

### Currently list manager uses this interface to populate List available to send to 

Screen Shot 2022-03-15 at 2 18 44 PM

That in turn populates 

The Bulk send dropdown
Screen Shot 2022-03-15 at 2 24 08 PM

And the subscribe block dropdown

Screen Shot 2022-03-15 at 2 25 34 PM

### DIFF
--- a/wordpress/wp-content/plugins/cds-base/classes/Modules/ListManager/app/src/store/SaveListData.ts
+++ b/wordpress/wp-content/plugins/cds-base/classes/Modules/ListManager/app/src/store/SaveListData.ts
@@ -1,0 +1,26 @@
+export const sendListData = async (
+  endpoint: string | undefined,
+  rest_nonce: string | undefined,
+  data: {}
+) => {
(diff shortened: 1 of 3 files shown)


pcraig3 (approved):
Approved! Let's keep working on this.

PR #632, opened 2022-03-16
Add URL preview component

pcraig3:
# Summary | Résumé

This PR actually includes everything in #629, but also adds a URL preview indicator to the form. Check out the latest commit to see the code for this preview by itself.

The URL preview widget sits underneath the "Title of your site" field on the signup form and it converts whatever is typed in into something that could be in a URL. 
- lowercase everything
- no apostrophes
- accents are removed

I added an `aria-live` role to the preview text container, and an `aria-describedby` attribute to the input to point at the preview container. I think those two together should make for a reasonable accessible solution to this.

## gif

Screen Recording 2022-03-16 at 14 40 14

## Accessibility

Tried it out in voice over and it seems like it actually works pretty well with "aria-live" added.

Screen Shot 2022-03-16 at 15 53 15

### DIFF
--- a/wordpress/wp-content/plugins/cds-base/classes/Modules/Forms/RequestSite/RequestSiteForm.php
+++ b/wordpress/wp-content/plugins/cds-base/classes/Modules/Forms/RequestSite/RequestSiteForm.php
@@ -153,12 +153,27 @@ class RequestSiteForm
                 ); ?>
             
                 <!-- start site -->
-                <?php Utils::textField(
-                    id: 'site',
-                    label: __('English title of your site', 'cds-snc'),
-                    description: __('This title will appear at the top of your site. You can change this later.', 'cds-snc'),
-                    value: $all_values['site']
-                ); ?>
+                <div class="focus-group">
+                    <label class="gc-label" for="site" id="site-label">
+                        <?php _e('English title of your site', 'cds-snc'); ?>
+                    </label>
+                    <div id="site-desc" class="gc-description" data-testid="description">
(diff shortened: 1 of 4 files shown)


wordpress/wp-content/plugins/cds-base/classes/Modules/Forms/RequestSite/RequestSiteForm.php
```diff
@@ -155,12 +153,29 @@ public function render($atts, $content = null): string
...
+                        class="gc-input-text" 
+                        id="site" 
+                        name="site" 
+                        value="<?php echo $all_values['site']; ?>"
+                        aria-describedby="url-typer"
+                        required
+                    />
+                    <div id="url-typer" class="url-typer gc-description" aria-live="polite" aria-atomic="true">
+                        <div class="url-typer--empty"><?php _e('Enter a title to preview your URL'); ?></div>
+                        <div class="url-typer--message displayNone"><?php _e('Your url will be:'); ?> <strong>articles.alpha.canada.ca/<span id="url-typer__preview"></span></strong></div>
```
timarney, line 174: Thinking we should update this to say

Your requested url will be --- or something like that .... in case the url is already taken.

timarney:
This is great --- nice work.

wordpress/wp-content/plugins/cds-base/classes/Modules/Forms/RequestSite/RequestSiteForm.php
pcraig3: Changed it to say "Your URL preview:", pretty close to what we are saying in the 'empty' message.

timarney (approved):

PR #635, opened 2022-03-18
EN / FR - Update css to draw focus 

timarney:
Making some small adjustments to see if that moves the needle on these tickets.

Suggesting we make these tweaks and see how it goes vs making larger changes to Core WP and WPML.

### Menus

> Language settings in screen shot need more prominence
https://github.com/cds-snc/gc-articles-issues/issues/276

menu

https://github.com/cds-snc/gc-articles-issues/issues/277

- Removed the "draft filters etc..." can re-add but for now I think they're distracting.

Screen Shot 2022-03-18 at 3 48 45 PM

### DIFF
--- a/wordpress/wp-content/plugins/cds-base/classes/Modules/Cleanup/css/admin.css
+++ b/wordpress/wp-content/plugins/cds-base/classes/Modules/Cleanup/css/admin.css
@@ -350,3 +350,30 @@ th.column-disable_user_login {
 .field-xfn, .field-description{
   display:none;
 }
+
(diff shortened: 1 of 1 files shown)


pcraig3:
Why remove the "all" option for the pages/posts? I think it's useful to see them all at the same time.

| before | after |
|--------|-------|
|   Screen Shot 2022-03-21 at 09 59 40   |   Screen Shot 2022-03-21 at 09 59 12  |

pcraig3 (approved):
👍  let's ship it

PR #637, opened 2022-03-21
Add detect to add tags

timarney:
- Adds a merge tags function


pcraig3 (approved):
Looks good! Hopefully this is the last change we ever have to make.

PR #638, opened 2022-03-21
Hide "Canada.ca" menu + unassign custom menu if exists

pcraig3:
# Summary | Résumé

This PR adds a new setting option to hide the Canada.ca menu.

It was a little tricky figuring out what hook to use, but it seems like it works now.

### Steps to test this:

- Look your default setup (Canada.ca menu should be there)
- Click "Hide Canada.ca menu" in the site settings. You should see the blue line only.
- Add your own custom menu to the header (it should appear, and then in the settings the option to "show the menu" should be selected)
- Click "Hide Canada.ca menu" in the site settings. Your custom menu should be gone. If you visit "Appearance > Menus > Locations", there should be no menu assigned to the header.

## Screenshots

| default page | page with only canada.ca menu removed | page with all header elements removed | page with custom menu |
|--------------|---------------------------------------|---------------------------------------|-----------------------|
|      Screen Shot 2022-03-21 at 14 47 13     |            Screen Shot 2022-03-21 at 14 47 29                   |  Screen Shot 2022-03-21 at 14 47 45                                  |         Screen Shot 2022-03-21 at 14 47 58          |

## Notes

Essentially, we have 3 different states we can be in:
1. Canada.ca top menu
2. We have our own top menu that we set
3. We don't want a top menu

In this PR, we are going to assume
- someone who clicks option (3) wants to get rid of _any_ top menu they have assigned (whether 1 OR 2) 
- someone who assigns a top menu (option 2) wants to unset option (3) 

What we want to avoid:
- someone builds their own menu but is not able to ever see it because they previously selected "hide canada.ca top menu" (in this case, it keeps getting programatically unassigned)
- Someone sets "hide canada.ca top menu" and then it still still shows their custom menu that they don't want

### DIFF
--- a/wordpress/wp-content/plugins/cds-base/classes/Modules/Blocks/src/styles/editor.css
+++ b/wordpress/wp-content/plugins/cds-base/classes/Modules/Blocks/src/styles/editor.css
@@ -235,3 +235,12 @@ width: 80px;
 #list-manager-app input.button[type="submit"] {
   margin-right: 10px;
 }
+
+.settings-warning {
+  display: inline-block;
+  border-radius: 4px;
+  border: 1px solid #dba617;
+  background: rgba(219, 166, 23, .3);
+  margin-bottom: 5px;
+  padding: 2px;
+}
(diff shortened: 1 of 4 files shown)


timarney:
Nice - works great.

timarney (approved):

PR #643, opened 2022-03-22
Cleanup old compiled .json translation files before compiling new ones

dsamojlenko:
# Summary | Résumé

Every time we run our compile-translations command, it generates a bunch of .json files with random strings in the filename. This means that they accumulate over time with a bunch of stale files hanging around.

This simply deletes previous .json files before generating new ones.


pcraig3 (approved):
✂️

PR #662, opened 2022-04-01
Add sign in + sign out links

timarney:
# Summary | Résumé

Fixes https://github.com/cds-snc/gc-articles-issues/issues/289

Visit a test site --- note the sign out or sign in links in the header next to the language switcher.


dsamojlenko (approved):
approved with my small change

PR #672, opened 2022-04-04
Allow email addresses ending with "onmicrosoft.com"

pcraig3:
# Summary | Résumé

Several departments are using these for their Azure landing zones. PSPC uses it, and ESDC was using it too. We will probably see this come up from a few departments.

Note that the `$goodEmails`, `$badEmails` variables in the test file weren't getting used, so I deleted them.

Resolves this ticket: https://github.com/cds-snc/gc-articles-issues/issues/297

### DIFF
--- a/wordpress/wp-content/plugins/cds-base/classes/Modules/Users/EmailDomains.php
+++ b/wordpress/wp-content/plugins/cds-base/classes/Modules/Users/EmailDomains.php
@@ -8,8 +8,7 @@ use WP_Error;
 
 class EmailDomains
 {
-    public const ALLOWED_EMAIL_DOMAINS = ['cds-snc.ca', 'gc.ca', 'canada.ca'];
-
+    public const ALLOWED_EMAIL_DOMAINS = ['cds-snc.ca', 'gc.ca', 'canada.ca', 'onmicrosoft.com'];
 
     public function __construct()
     {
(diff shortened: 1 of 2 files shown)


andrewleith:
Not sure I should comment on this given its my first day and I don't have much context, but I just wanted to mention that `onmicrosoft.com` is pretty broad. Anyone (including people not inside the GOC) can get a `sub-domain.onmicrosoft.com`.  PSPC's inno lab in particular is `pspcinnovation.onmicrosoft.com`.

Depending on what you're going for here, you may need to whitelist more specific domains. :)

pcraig3:
Hi Andrew, 

Yeah, it's true that it's pretty broad. It allows other MS emails, as well as anyone who buys a similar domain ("paul@lisbonmicrosoft.com"). We also do this with our other email suffixes, allowing people to sign up with potentially weird emails by not enforcing subdomains on them ("paul@ohcanada.ca"). 

For this PR in particular, it seems like we have 3 options:

1. only allow this PSPC domain
2. subdomain to 'onmicrosoft' (add a dot to the beginning)
3. keep what we have now, allowing a pretty broad allowable list

I don't have a huge preference either way, but I think 3 is okay. I think that given our "add new email to allowed email list" process is sufficiently slow and our creating sites is still a manual process, it's fine to just default to a wider selector and then review them one by one.

timarney:
@willeybryan --- any preference on this?

On my end I would prefer to limit and widen as needed.  

That said per @pcraig3 

> creating sites is still a manual process

So I think we'll end up validating either way 

--- if they can't submit they'll likely contact us 
--- if they can submit we'll likely contact them to validate

pcraig3:
Okay, allowing subdomains of "onmicrosoft.com" by adding a dot (.). Eg, these will no longer work:

- `paul@onmicrosoft.com`
- `paul@lisbonmicrsoft.com`

timarney (approved):

PR #674, opened 2022-04-04
Update list manager to work for local dev outside of WP

timarney:
Updating so we can do local dev for List Manager outside of WP 

This adds a simple json file vs hitting the list manager API --- also pulls in WP styles.

This isn't intended to be a 1:1 match for look or functionality just easier app updating.  

To test

From the ListManager app directory 

```
npm run start
```

View your http:localhost:3000

Screen Shot 2022-04-04 at 10 18 29 AM

<hr>

Once this is merged we can look to upgrade to React 18

### DIFF
--- a/wordpress/wp-content/plugins/cds-base/classes/Modules/ListManager/app/src/components/ListForm.tsx
+++ b/wordpress/wp-content/plugins/cds-base/classes/Modules/ListManager/app/src/components/ListForm.tsx
@@ -34,93 +34,95 @@ export const ListForm = ({ handler, formData = {}, serverErrors = [] }: { handle
         <form onSubmit={handleSubmit(handler)}>
             <input id="service_id" type="hidden" {...register("service_id", { required: true })} />
             <table id="form-table" className="form-table">
-                <tr>
-                    <th><label className="required" htmlFor="name"><Asterisk />{__("Name", "cds-snc")}</label></th>
-                    <td>
-                        <div className={errors.name ? "error-wrapper" : ""}>
-                            {errors.name && <span className="validation-error">{errors.name?.message || __("Name is required", "cds-snc")}</span>}
-                            <input id="name" style={textWidth} type="text" {...register("name", { required: true })} />
-                        </div>
-                    </td>
-                </tr>
-                <tr>
-                    <th><label className="required" htmlFor="language"><Asterisk />{__("Language", "cds-snc")}</label></th>
-                    <td>
-                        <div className={errors.language ? "error-wrapper" : ""}>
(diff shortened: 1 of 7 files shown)


wordpress/wp-content/plugins/cds-base/classes/Modules/ListManager/app/src/components/ListForm.tsx
```diff
@@ -34,93 +34,95 @@ export const ListForm = ({ handler, formData = {}, serverErrors = [] }: { handle
...
-                    <th><label htmlFor="subscribe_email_template_id">{__("Subscribe template id", "cds-snc")}</label></th>
-                    <td>
-                        <div className={errors.subscribe_email_template_id ? "error-wrapper" : ""}>
-                            {errors.subscribe_email_template_id && <span className="validation-error">{errors.subscribe_email_template_id?.message}</span>}
-                            <input id="subscribe_email_template_id" style={textWidth} type="text" {...register("subscribe_email_template_id")} />
-                            <div className="role-desc description">
-                                <details>
-                                    <summary>{__("See example template ID format.", "cds-snc")}</summary><code>ex4mp1e0-d248-4661-a3d6-0647167e3720</code>
-                                </details>
+                <tbody>
```
pcraig3, line 37: Is the addition of the `<tbody>` the only new thing in this file? (other than all the indenting? The diff is hard to read.)

timarney: Ya there was some lint errors with `<tbody>` + changed to `tds` as they weren't in the table header

pcraig3: You can leave them as headers, just use the `scope="row"` attribute if you are doing so. It's pretty common to have tables with "header" cells at the beginning of the row. (I have been learning this recently.)

for example, this table would have headers at the top and for the first cell in the row

```
|        | price | is it healthy? | is it good? |
|--------|-------|----------------|-------------|
| coffee | $1.00 | sort of        | yes         |
| water  | $.50  | yes            | no          |
| beer   | $1.50 | not really     | yes         |
```

Here's an example: https://www.w3.org/WAI/tutorials/tables/two-headers/

pcraig3 (approved):
Pretty neat, gonna pre-approve.

A couple things:

- you should be good to keep using the `th`s as row headers (see comment)
- the list import fails to exit gracefully (it just kind of hangs) and there's no "back" if you do click it, but this is kind of a "pre" interface anyway.

PR #677, opened 2022-04-04
List manager - React 18 --- requires #674

timarney:
This is a follow-up after https://github.com/cds-snc/gc-articles/pull/674


pcraig3 (approved):
It's working for me 👍

PR #683, opened 2022-04-05
Add site url to welcome email

timarney:
Adds ((site_url)) variable for existing users 

When sending the Welcome email we don't want to use the password reset link for existing users.  This adds a site_url variable for that use case.

Fixes 

https://github.com/cds-snc/gc-articles-issues/issues/298

### DIFF
--- a/wordpress/wp-content/plugins/cds-base/classes/Modules/Users/Users.php
+++ b/wordpress/wp-content/plugins/cds-base/classes/Modules/Users/Users.php
@@ -196,8 +196,7 @@ class Users
             $password_message = $uniqueUrl;
         }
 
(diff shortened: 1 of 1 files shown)


pcraig3 (approved):
All good!

Tried it with 4 different cases and all the emails looked good.

|               | default template | "welcome" template |
|---------------|------------------|--------------------|
| new user      |    Screen Shot 2022-04-05 at 14 42 17        |         Screen Shot 2022-04-05 at 14 42 02           |
| existing user |    Screen Shot 2022-04-05 at 14 41 44     |    (first link is different)    Screen Shot 2022-04-05 at 14 46 19     |

PR #691, opened 2022-04-05
Add policy link text field for Subscribe

timarney:
Fixes https://github.com/cds-snc/gc-articles-issues/issues/285


pcraig3 (approved):
Works well! Once we add this, the form isn't bilingual anymore (the string "Privacy Policy" doesn't translate), but we can update that as a separate PR.

PR #694, opened 2022-04-06
Add checklists permission

pcraig3:
# Summary | Résumé

Adds the `manage_checklists` permissions when checklist plugin is active, removes them when the plugin is deactivated.

Took me a while to figure out because we also filter menu items. (So even when the permission was working, the menu page was not showing up.) Eventually figured it out, but ended up going down a few dead ends.


timarney (approved):
If / when we auto activate the plugin we should write some tests for this functionality.

PR #699, opened 2022-04-07
Update WP 5.9.3 + PHP 8.1

timarney:
Closes https://github.com/cds-snc/gc-articles-issues/issues/311


dsamojlenko:
Should also update the Dockerfile.local for our local environments

PR #702, opened 2022-04-11
Remove comments column

timarney:
Fixes https://github.com/cds-snc/gc-articles-issues/issues/303

Adds filter to remove column

Screen Shot 2022-04-11 at 1 28 58 PM


pcraig3 (approved):
👍 
All works!

I wonder if we want to remove some of those Yoast columns as well eventually.

timarney:
@pcraig3 

For now --- we'll remove those via the Yoast settings toogle but yes 

See https://github.com/cds-snc/gc-articles-issues/issues/304#issuecomment-1089063107

PR #707, opened 2022-04-13
Add a modal popover to make the Checklists interactions more obvious

pcraig3:
# Summary | Résumé

This PR adds a bunch of jQuery which adds modals to our checklist feature. Essentially, if you try to publish or update a post without doing all the checks, you get a modal which prevents you from publishing (if required checks are not met) or warns you away from publishing (if only recommended checks are met).

It introduces (basically) a 4-outcome flowchart:

Try to publish/update:

1. All checks met: Publish/update like normal
2. All "required" checks met: Warning modal, but then Publish/update anyway
3. All "required" checks met: Warning modal, decide not to Publish/update, focus the checklist
4. Not all "required" checks met: Prevent publishing modal, focus the checklist

## gif

Screen Recording 2022-04-20 at 18 27 32

## Notes:

- Most of the original code comes from another plugin, the other file is here: https://github.com/brainstormforce/pre-publish-checklist/blob/master/assets/js/ppc-checkbox.js
- Essentially what is happening is that two hidden modals are added to editing pages, and then the "publish/update" buttons are replaced with custom publish.../update... buttons. Clicking the custom buttons pops up the modals. One modal prevents publishing (if required checks are not met) and one modal 'warns' the user before publishing (if "recommended" checks are not met). Once all checks are met, the regular WP Update/Publish buttons are used and the modals never come up.
- The `ppc_checkbox_function` function does most of the work of making sure the interface is in sync with the data. It runs whenever the checklist items are changed (manually or automatically) and when the post/page status (draft/published) changes.
- The JS file relies on specific classnames for the WordPress Publish/Update buttons, the Modals, and the custom buttons. If those change (which isn't too likely, but not impossible), then the JS is going to break.
- The original English in modals isn't changed in this PR, we can do that in a future PR.
   - The text of the custom publish/update buttons in the JS file also needs to be translated
- I raised an issue in the PublishPress plugin for this functionality, so it may eventually show up: https://github.com/publishpress/PublishPress-Checklists/issues/353

### DIFF
--- a/wordpress/wp-content/plugins/cds-base/classes/Modules/Checklists/js/meta-box.js
+++ b/wordpress/wp-content/plugins/cds-base/classes/Modules/Checklists/js/meta-box.js
@@ -1,0 +1,229 @@
+jQuery(document).ready(
+    function() {
+        /* Assigning vars */
+        var ppc_publish_flag = 0;
+        var ppc_checkboxes = jQuery('.ppc_checkboxes[type="checkbox"]');
+        var ppc_checkboxes_length = jQuery('.ppc_checkboxes[type="checkbox"]').length;
+        var countCheckedppc_checkboxes = ppc_checkboxes.filter(':checked').length;
+        var ppc_percentage_completed = (countCheckedppc_checkboxes / ppc_checkboxes_length) * 100;
+        jQuery('.ppc-percentage').attr('style', 'width:' + ppc_percentage_completed + '%');
+        jQuery(".ppc-percentage-value").html(Math.round(ppc_percentage_completed) + "%");
+
+
+        // .editor-post-publish-panel__toggle   = "publish" button for unpublished posts 
+        // .editor-post-publish-button          = "update" button for already-published posts
+
+        //function to be executed when the itemlist changes.
+        
+        var ppc_checkbox_function = function() {
+            var ppc_checkboxes = jQuery('.ppc_checkboxes[type="checkbox"]');
+            var ppc_checkboxes_length = jQuery('.ppc_checkboxes[type="checkbox"]').length;
+            var countCheckedppc_checkboxes = ppc_checkboxes.filter(':checked').length;
+            if (ppc_checkboxes_length == countCheckedppc_checkboxes) { // if all the checkboxes are checked (lets publish!!)
+                if (jQuery('.editor-post-publish-panel__toggle').length == 1) {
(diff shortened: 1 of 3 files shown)


wordpress/wp-content/plugins/cds-base/classes/Modules/Checklists/js/meta-box.js
```diff
@@ -0,0 +1,229 @@
...
+
+
+        // .editor-post-publish-panel__toggle   = "publish" button for unpublished posts 
+        // .editor-post-publish-button          = "update" button for already-published posts
+
+        //function to be executed when the itemlist changes.
+        
+        var ppc_checkbox_function = function() {
+            var ppc_checkboxes = jQuery('.ppc_checkboxes[type="checkbox"]');
+            var ppc_checkboxes_length = jQuery('.ppc_checkboxes[type="checkbox"]').length;
```
timarney, line 20: Do we need these if they are assigned above

wordpress/wp-content/plugins/cds-base/classes/Modules/Checklists/js/meta-box.js
```diff
@@ -0,0 +1,229 @@
+jQuery(document).ready(
```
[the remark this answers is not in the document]
pcraig3: Refactored the jQuery pretty heavily, swapped the `jQuery` for `$`

wordpress/wp-content/plugins/cds-base/classes/Modules/Checklists/js/meta-box.js
pcraig3: No!

pcraig3:
> For a follow up PR
> 
> I think something like this will work to monitor the post status
> 
> ```js
> const getPostStatus = () => wp.data.select('core/editor').getEditedPostAttribute('status');
> 
> let postStatus = getPostStatus ();
> 
> wp.data.subscribe(() => {
>     const newPostStatus = getPostStatus();    
>     if( postStatus !== newPostStatus ) {
>         console.log(`it changed to ${newPostStatus}`);
>     }
>     postStatus = newPostStatus;
> });
> ```

Added in this commit: https://github.com/cds-snc/gc-articles/pull/707/commits/54babd9e2af13822f3cfc6acc3f1419d9fa5939e

timarney:
Works great.

timarney (approved):

timarney (approved):

PR #708, opened 2022-04-13
Add bulk send sms capability

timarney:
Adds a bulk send SMS capability under user profile for Super Admins 

Screen Shot 2022-04-13 at 9 36 58 AM

Note: I also renamed List Manager to GC Lists

### DIFF
--- a/wordpress/wp-content/plugins/cds-base/classes/Modules/Notify/ListManagerUserProfile.php
+++ b/wordpress/wp-content/plugins/cds-base/classes/Modules/Notify/ListManagerUserProfile.php
@@ -79,7 +79,7 @@ class ListManagerUserProfile
 
         $user = new WP_User($userId);
         $user->remove_cap('list_manager_bulk_send');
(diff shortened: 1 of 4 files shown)


wordpress/wp-content/plugins/cds-base/classes/Modules/Notify/ListManagerUserProfile.php
```diff
@@ -58,9 +79,14 @@ public function updateListManagerMeta($userId)
 
         $user = new WP_User($userId);
         $user->remove_cap('list_manager_bulk_send');
+        $user->remove_cap('list_manager_bulk_send_send');
```
pcraig3, line 82: This is probably `list_manager_bulk_send_sms`, not `list_manager_bulk_send_send`

pcraig3 (commented):
Just the one comment, but otherwise this looks good.

pcraig3 (approved):

PR #717, opened 2022-04-21
Fix list id issue

timarney:
GC Lists - list id is coming back as undefined when trying to edit a list.

This uses the original row data vs the row `resolved` data

https://react-table.tanstack.com/docs/api/useTable


pcraig3 (approved):
👍  works!

PR #721, opened 2022-04-21
Fix login link on 404 page

dsamojlenko:
# Summary | Résumé

See cds-snc/gc-articles-issues#321

Fixes an issue where the Login link in the header of a 404 page would simply link to `#`.

The root cause was this bit of code in the WPS Hide Login plugin:

```
if ( is_404() ) {
	return '#';
}
```

Assuming the intention here is to not accidentally leak the "hidden" login url from the 404 page. We have decided to include these Sign in links deliberately, and so they should work as intended.

The plugin sets the login_url through a WP filter. This PR overrides that filter with our own which contains most of the same code, replacing the `is_404` bit with our own logic.

## To test
- Setup one or more sites on your localhost
- Note that the default url for WPS Hide Login is `/login` ... you can change this in settings
- Set a login URL for each site you are testing (they can all be the same)
- Logout and visit a page that would trigger a 404, ie `/xxx` or `/my-subsite/xxx`
- Try the Sign in link in the header - it should take you to the appropriate sign in link


timarney (approved):
Nice

PR #723, opened 2022-04-22
Add Table of contents pattern

timarney:
Adds a Table of contents pattern as requested in https://github.com/cds-snc/gc-articles-issues/issues/327 .

This is more of a proof of concept vs finished code but I think it's worth adding.

Noting there's no interface to add a class to one of the `<a>` tags for the active state but there might be at some point see

https://github.com/WordPress/gutenberg/issues/23011

With active class applied to the  `<a>`  using the code editor
Screen Shot 2022-04-22 at 10 10 42 AM

### DIFF
--- a/wordpress/wp-content/themes/cds-default/functions.php
+++ b/wordpress/wp-content/themes/cds-default/functions.php
@@ -220,6 +220,11 @@ if (! function_exists('cds_register_block_patterns')) :
                 'categories'    => array( 'homepage' ),
                 'content'       => cds_get_block_pattern_markup('homepage/landing'),
             ),
+            'cds/toc' => array(
+                'title'         => esc_html__('Table of contents', 'cds-snc'),
+                'categories'    => array( 'homepage' ),
+                'content'       => cds_get_block_pattern_markup('homepage/toc'),
+            ),
         ));
 
         // Register block patterns.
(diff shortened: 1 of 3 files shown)


pcraig3:
Looks good!

Is there a way of adding a class based on the current URL path? The example is kind of broken, but it shows adding the `active` class to the current element (it makes it dark blue).

I tried it out locally, just manually adding the class for the page I was on.

Screen Shot 2022-04-22 at 15 44 29

I was also going to suggest that this be marked up as some kind of `nav` element, but I checked the WET implementation and they aren't doing it, so I guess we're safe.

pcraig3:
Also, it seems weird they use an unordered list when their example seems to specifically number the options.

pcraig3:
Also, there is another "Table of Contents" Yoast block. If someone tries to use this one, they might end up clicking the wrong one (just something to note).

Screen Shot 2022-04-22 at 16 06 10

timarney:
> Also, it seems weird they use an unordered list when their example seems to specifically number the options.

Agreed

Maybe we can remove the Yoast block in another PR

It's a bit odd --- would need to look at it further but it just adds a heading

```
<!-- wp:yoast-seo/table-of-contents -->
<div class="wp-block-yoast-seo-table-of-contents yoast-table-of-contents"><h2>Table of contents<br>kjjkj</h2></div>
<!-- /wp:yoast-seo/table-of-contents -->
```

timarney:
Updated to filter out yoast-seo/table-of-contents

dsamojlenko (approved):

PR #737, opened 2022-04-28
Intranet improvements

pcraig3:
# Summary | Résumé

This PR adds a couple of small CSS fixes to improve the CDS intranet site that we now have up and running.

1. Fix the search button in the search bar: align it so that it sits inside the search input
2. Add padding to divs with an explicit background colour: they are using it like a panel and currently it looks terrible with no padding

### Fix for search button 

| before | after |
|--------|-------|
|  Search button not aligned with the search input      |   Now it is    |
|  Screen Shot 2022-04-28 at 16 27 50    |   Screen Shot 2022-04-28 at 16 27 43  |

### More padding for panels

| before | after |
|--------|-------|
|  Text is crowded in div with background colour     |   Now it has more padding  |
|  Screen Shot 2022-04-28 at 16 39 24    |  Screen Shot 2022-04-28 at 16 39 20  |


timarney (approved):
Nice

PR #743, opened 2022-05-04
Allow GC Admins, Editors, and Writers to embed iFrames

pcraig3:
# Summary | Résumé

This PR adds the `unfiltered_html` capability to users who have been approved by a superadmin. Here's what the screenshot looks like:

Screen Shot 2022-05-11 at 18 57 31

Unlike our other capabilities, which we can just add to roles during the setup and they will just work, `unfiltered_html` is treated differently.

> User roles other than Super Admin cannot be assigned the `unfiltered_html` capability in WordPress Multisite. You can set it, but then WordPress disables the capability after the fact.

Because of this, we have to add 2 capabilities here:
- add the 'unfiltered_html' capability to select users
- add a custom `allow_unfiltered_html` permission, which we check in the 'map_meta_cap' filter.

Adding `unfiltered_html` to the roles array on its own does not work, but this does.

Resolves this issue: https://github.com/cds-snc/gc-articles-issues/issues/338

Sources:
- https://smartslider.helpscoutdocs.com/article/1983-how-to-give-access-to-smart-slider-for-non-admin-users#multisite
- https://hegeman.me/kb/wordpress-multisite-add-unfiltered-html-capability-to-role/

### DIFF
--- a/wordpress/wp-content/plugins/cds-base/classes/Modules/Cleanup/Profile.php
+++ b/wordpress/wp-content/plugins/cds-base/classes/Modules/Cleanup/Profile.php
@@ -15,6 +15,9 @@ class Profile
         add_action('wpml_user_profile_options', [$this, 'wpmlOptions']);
         add_action('additional_capabilities_display', [$this, 'yoastOptions']);
 
+        add_action('edit_user_profile', [$this, 'displayUnfilteredHTMLMeta']);
+        add_action('edit_user_profile_update', [$this,'updateUnfilteredHTMLMeta']);
+
         if (is_admin()) {
             remove_action('admin_color_scheme_picker', 'admin_color_scheme_picker');
             remove_action('personal_options', 'wpml_show_user_options');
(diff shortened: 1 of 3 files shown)


timarney:
@willeybryan 

Allowing this for all users does open up a wider surface area for attack. 

via https://wordpress.org/plugins/unfiltered-mu

> Warning! This is a very dangerous plugin to activate if you have untrusted users on your site. Any user could add Javascript code to steal the login cookies of any visitor who runs a blog on the same site. The rogue user can then inpersonate any of those users and wreak havoc. 

> If you use this plugin your site will be hacked in one way or another if you allow anonymous users on the Internet to create blogs on your site. It’s very dangerous.

> Are you still 100% sure you want to use this plugin?

Thinking we might want to add a permission and open this up for specific users if requested.  Thoughts.  

If not ... we should really look at enforcing two factor and other types of detection to limit potential attacks.

willeybryan:
That totally makes sense. I assumed that since we were only allowing GC Staff to connect and not untrusted users the attack surface increase would be minimal. There are a lot of staff in government however and a lot of accounts that could be compromised without us knowing. If the attack surface area gives you cause for concern then we should probably do something about it, I know I can be pretty lax about these things. 

Enforcing two factor is a good idea in general as we develop the software more. I am going to make a ticket for this anyway as it's a security control. 

If we are trying to fit it in with the role based access controls (RBAC), we could make it available to GC Admins only and maybe in some future state a GC Developer role. 

Any other ideas?

timarney (approved):

PR #744, opened 2022-05-04
Hide login/out link on subsites

dsamojlenko:
# Summary | Résumé

Hides the Sign in/out link when on subsites.

Closes cds-snc/gc-articles-issues#334


pcraig3 (approved):
👍

pcraig3:
Should this be a "site setting"? It seems like something some people might want to have.

timarney:
> Should this be a "site setting"? It seems like something some people might want to have.

I have a feeling if folks are sending links to others they might expect to have a login link ... that said I think we can merge this and await feedback from users.

PR #761, opened 2022-05-13
Add "Last modified" date to all posts and pages

pcraig3:
# Summary | Résumé

We saw one of our sites creating this manually. This is a required Canada.ca element, so we can just include it for all posts and pages. Adding this to the CDS Default theme. Don't need it for the CDS Redirector theme because that is our "headless CMS" use case, so they wouldn't use the default templates in that case.

Resolves: https://github.com/cds-snc/gc-articles-issues/issues/346

Source:
- https://design.canada.ca/common-design-patterns/date-modified.html

## Screenshot

Screen Shot 2022-05-13 at 15 33 45

### DIFF
--- a/wordpress/wp-content/themes/cds-default/inc/template-functions.php
+++ b/wordpress/wp-content/themes/cds-default/inc/template-functions.php
@@ -431,3 +431,17 @@ function get_top_nav(): string
 
     return '';
 }
+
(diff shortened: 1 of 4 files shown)


timarney (approved):

pcraig3:
Gonna merge!

PR #763, opened 2022-05-16
Create a 2-page contact form

pcraig3:
# Summary | Résumé

Since we started to get spammed on the Contact form, we are splitting it into 2 pages. The result will be that's easy for humans to fill out but hard for bots.

The majority of this code is brought over from the "Request a site" form, which is already broken into 2 pages.

It's not the most beautiful code in the world but it's been working okay for what we need it for.

Resolves: https://github.com/cds-snc/gc-articles-issues/issues/364

## Notes

- Updated the "group" divs with an id so that clicking the error messages takes you to that question
- Updated the ids of the optional fields because it seemed too confusing to have a div wrapping the optional fields with an id of `optional-usage` and the field itself with an id of `usage-optional`
- Fixed a pretty minor axe problem where some of the description ids on the contact form were duplicated

## Screenshots

| before | after (page 1) | after (page 2) |
|--------|----------------|----------------|
|  all fields on one page      |     page 1: what is your message?           |      page 2: who are you?          |
|   localhost_contact-form_ (1)   |     localhost_contact-form_        |     localhost_contact-form           |

### DIFF
--- a/wordpress/wp-content/plugins/cds-base/classes/Modules/Blocks/src/styles/public.css
+++ b/wordpress/wp-content/plugins/cds-base/classes/Modules/Blocks/src/styles/public.css
@@ -123,6 +123,12 @@ form * {
   margin-top: .75em;
 }
 
+form div[role="group"][id]:focus {
+  outline: 2px solid blue;
+  outline-offset: 5px;
+  border-radius: 2px;
+}
+
 form .focus-group {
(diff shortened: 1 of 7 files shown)


timarney (approved):
Going to approve and we can test further on staging ... 

Testing locally - think there might be a bug but could be my setup

PR #803, opened 2022-05-31
Refactor gc-lists templates to simplify by moving logic to Menu.php

dsamojlenko:
# Summary | Résumé

Tidied up the templates used to render the lists and messages apps:
- Moved logic out to Menu.php
- Added phpdoc headers so phpstorm doesn't complain about Undefined variables

### DIFF
--- a/wordpress/wp-content/plugins/gc-lists/src/Menu.php
+++ b/wordpress/wp-content/plugins/gc-lists/src/Menu.php
@@ -4,6 +4,8 @@ declare(strict_types=1);
 
 namespace GCLists;
 
+use CDS\Modules\Notify\Utils;
+
 class Menu
 {
     protected static $instance;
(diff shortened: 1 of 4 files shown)


wordpress/wp-content/plugins/gc-lists/src/Menu.php
```diff
@@ -61,20 +61,75 @@ public function addSubscriberListsSubmenuItem()
...
+     * Get services array
+     *
+     * @return array
+     */
+    public function getServices()
+    {
+        return [
+            'name' => __('Your Lists', 'cds-snc'),
+            'service_id' => $this->getServiceId(),
+            'sendingTemplate' => get_option('NOTIFY_GENERIC_TEMPLATE_ID', '')
```
timarney, line 95: I don't think we need the sendingTemplate here.

PR #875, opened 2022-06-27
feat: toggleable waf

mohamed-cds:
Adding the ability to enable/disable the WAF. The first use case is for when security assessments are being performed. This may also come in handy in the future if there's a production incident related to blocked requests. Easily setting all the rules to `count` using a feature flag will reduce the risk of error.

### DIFF
--- a/infrastructure/terragrunt/aws/load-balancer/waf.tf
+++ b/infrastructure/terragrunt/aws/load-balancer/waf.tf
@@ -23,7 +23,7 @@ resource "aws_wafv2_web_acl" "wordpress_waf" {
 
     override_action {
       dynamic "none" {
-        for_each = var.enable_waf ? [""] : []
+        for_each = var.enable_waf == true ? [""] : []
         content {
         }
       }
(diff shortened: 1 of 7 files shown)


infrastructure/terragrunt/aws/load-balancer/waf.tf
```diff
@@ -22,7 +22,17 @@ resource "aws_wafv2_web_acl" "wordpress_waf" {
...
+        for_each = var.enable_waf ? [""] : []
+        content {
+        }
+      }
+
+      dynamic "count" {
+        for_each = var.enable_waf == false ? [""] : []
+        content {
+        }
+      }
```
patheard, line 35: I really like this pattern, but a minor comment - could we switch to be consistent in how we test for `true/false`?   I'm not fussed either way:

```hcl
      dynamic "none" {
        for_each = var.enable_waf == true ? [""] : []
        content {
        }
      }

      dynamic "count" {
        for_each = var.enable_waf == false ? [""] : []
        content {
        }
      }
``` 

or

```hcl
      dynamic "none" {
        for_each = var.enable_waf ? [""] : []
        content {
        }
      }

      dynamic "count" {
        for_each = !var.enable_waf ? [""] : []
        content {
        }
      }
```

patheard (commented):
Super handy to have the toggleable WAF.  

I'll leave the approval for @timarney or @dsamojlenko so they can weigh in on the change.

dsamojlenko (approved):

PR #913, opened 2022-07-07
Fail nonce check on WPML rest actions

dsamojlenko:
# Summary | Résumé

There are a bunch of WPML Rest Actions that are not multi-site safe. Since we don't use any of these features, we're just going to disable them by force-failing the nonce check.


timarney:
Screen Shot 2022-07-07 at 2 20 56 PM

PR #926, opened 2022-07-08
add super admin check

timarney:


pcraig3:
Does this fire for every nonce check (including an admin adding a new user, for example?) or is it just for some specific use case?

pcraig3 (approved):

dsamojlenko (approved):

PR #931, opened 2022-07-11
Remove cloudfront attached security header policies

dsamojlenko:
# Summary | Résumé

These headers are moving into a plugin in the application layer for easier management, so this PR removes the custom CloudFront headers and allow-lists the relevant headers from the web service.

Dependent on #932 
Also, once #932 is merged, the cds-security-headers plugin should be network-enabled in all environments.


dsamojlenko:
Setting to draft to prevent accidental merge. See note above about #932

pcraig3 (approved):
Seems to do what it says 👍

PR #932, opened 2022-07-11
Add more security headers to management plugin

dsamojlenko:
# Summary | Résumé

Fleshes out the security headers plugin with HSTS, XSS, X-Frame-Options, and X-Content-Type-Options headers.

Also conditionally adds only the CSP headers on non-admin pages.

### DIFF
--- a/wordpress/wp-content/plugins/cds-security-headers/cds-security-headers.php
+++ b/wordpress/wp-content/plugins/cds-security-headers/cds-security-headers.php
@@ -87,11 +87,16 @@ function cds_security_headers($headers)
         }
     }
 
-    $headers['Content-Security-Policy'] = $csp;
+    if (! is_admin()) {
+        $headers['Content-Security-Policy'] = $csp;
+    }
+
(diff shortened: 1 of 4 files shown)


wordpress/wp-content/plugins/cds-security-headers/cds-security-headers.php
```diff
@@ -87,11 +89,16 @@ function cds_security_headers($headers)
...
+    $headers['X-Frame-Options'] = 'SAMEORIGIN';
+    $headers['X-Content-Type-Options'] = 'nosniff';
 
     return $headers;
 }
 
-if (! is_admin()) {
-    add_filter('wp_headers', 'cds_security_headers');
-}
+add_filter('wp_headers', 'cds_security_headers');
```
pcraig3, line 104: This only seems to run in the frontend (not in WP admin screens), so the if condition above isn't needed.

pcraig3 (commented):
Would be good to add a line removing the "X-Powered-By" header as well maybe.

Screen Shot 2022-07-11 at 15 37 12

timarney (approved):

PR #959, opened 2022-07-15
Add "text messages" screens

pcraig3:
# Summary | Résumé

This PR adds a flow for choosing what message to send (email or text messages), but only for users who have the right permissions (eg, superusers and users with "Bulk Send SMS" capabilities).

Changes:
- Added `message_type` query param to filter draft templates (email-only users shouldn't see phone messages in drafts)
- Added "Choose message" screen
- Updated 'Edit message' screen to send emails or phone messages (no subject line in phone messages)
  - Also changed the URL from `/messages/edit/:id` to `messages/edit/:message_type/:id`
- When choosing lists to send to, only show lists with the same message_type
- Add "not authorized" component for users with insufficient permissions trying to visit phone screens
- Added updated French translations for new pages and some that I missed before

## gif

Screen Recording 2022-07-15 at 10 50 02


timarney (approved):

pcraig3:
Gonna merge, here we go.

PR #973, opened 2022-07-19
Sanitize input and escape output of custom Site Settings options

dsamojlenko:
# Summary | Résumé

Adds sanitization callbacks for all custom Site Settings options, and adds a few missing escape functions too.

### DIFF
--- a/wordpress/wp-content/plugins/cds-base/classes/Modules/Site/SiteSettings.php
+++ b/wordpress/wp-content/plugins/cds-base/classes/Modules/Site/SiteSettings.php
@@ -100,11 +100,7 @@ class SiteSettings
             'site_settings_group', // option_group
             'collection_mode_maintenance_page',
             function ($input) {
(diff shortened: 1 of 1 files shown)


wordpress/wp-content/plugins/cds-base/classes/Modules/Site/SiteSettings.php
```diff
@@ -101,72 +96,124 @@ public function collectionSettingsPageInit()
...
-        register_setting(
-            'site_settings_group', // option_group
-            'collection_mode',
-        );
-
         register_setting(
             'site_settings_group', // option_group
             'collection_mode_maintenance_page',
+            function ($input) {
+                if (is_int($input)) {
```
timarney, line 103: Thinking we should use `return intval($input);` here --- tested using the existing and it doesn't hold the right value

timarney (approved):

PR #986, opened 2022-07-26
Update list view

timarney:
# Summary | Résumé

Updates the list view to use the new Figma design
Screen Shot 2022-07-26 at 10 05 40 AM

**New design:**
Screen Shot 2022-07-26 at 10 04 20 AM


timarney:
I added some initial form styling to this PR as well --- it's not the final design but preps things for moving forward.

Screen Shot 2022-07-26 at 2 45 42 PM

pcraig3 (approved):
Approved! Looks like it's heading the right way. A couple observations though.

### Input elements are sometimes below, sometimes on the right side

On the "Choose message type" screen (as well as "Create message"), we are using the table-based "wp settings" layout where all the labels are on the left side and all the inputs are on the right side. When moving between lists and messages, you can see the layout changes. 

I prefer the one that's side by side and I think it's more what WP looks like, but we don't have to change it for this PR to go in.

| messages: radios are right of the label | lists: radios are below the label |
|--------|-------|
|  Screen Shot 2022-07-27 at 10 19 38  |  Screen Shot 2022-07-27 at 10 19 47   |

### Super minor: "Edit a list"

This is not something your PR caused, but I just noticed it while reviewing your PR. Since the "3 links" pattern is just inline, at certain screen sizes the text for "Send to a list" runs to the next line and it looks like it's part of the "Edit" link.

I don't think it matters until someone else notices it, but just FYI since we are using the same pattern on the lists.

| list of draft messages | 
|--------|
|   "Edit a list" is actually two different links.     |  
|  Screen Shot 2022-07-27 at 10 19 27   |

PR #988, opened 2022-07-26
Update references to alpha domain

dsamojlenko:
# Summary | Résumé

Updates references to alpha domain to new articles.canada.ca domain.


pcraig3 (commented):
This one too, right? 

https://github.com/cds-snc/gc-articles/blob/main/wordpress/wp-content/plugins/gc-lists/resources/js/src/lists/components/CreateList.tsx#L36-L42

PR #992, opened 2022-07-27
Add esc_html to text field

timarney:
Update to handle https://github.com/cds-snc/gc-articles-issues/issues/376#issuecomment-1195848765


pcraig3:
Interesting, I wouldn't have thought of this one either.

pcraig3 (approved):

PR #1000, opened 2022-07-27
Move notify settings

timarney:
# Summary 

Moves the Notify settings under the GC Lists menu

Noting: final design for this page isn't ready yet - but this moves things forward

Screen Shot 2022-07-27 at 4 11 32 PM

### DIFF
--- a/cypress/integration/notify.settings.spec.js
+++ b/cypress/integration/notify.settings.spec.js
@@ -1,4 +1,4 @@
-describe('Notify API Settings', () => {
+describe('GC Lists Settings', () => {
     before(() => {
       cy.testSetup();
     });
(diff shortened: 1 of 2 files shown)


pcraig3 (approved):
Does what it says 👍

pcraig3:
Why are all the headings different on those 3 pages?

1. Messages: h1, font weight 400
2. Subscriber lists: h1, font weight 700
3. Settings: h2, font weight 700

Probably we should make them all #1 or #2

timarney:
> Why are all the headings different on those 3 pages?

Ya we can clean up the pages etc... when we're doing the updates to the list form page.

PR #1012, opened 2022-08-02
Yoast breadcrumb - Home translation

timarney:
# Summary | Résumé

Ensures the "Home" link in the breadcrumb gets translated when navigating FR site.

The "Home" link is set under the Yoast search appearance settings and doesn't end up in a translation file.  Given we're not using the WPML string translation plugin this string ends up not being translated.

This PR adds a filter to ensure `Home` gets translated to `Accueil`

Screen Shot 2022-08-02 at 1 59 07 PM


pcraig3:
This works for me unless "Home" is the current page. If we're on another page, the "home" link is translated, but when it is not a link, it seems like it's not working.

pcraig3 (approved):
Seems good 👍

PR #1013, opened 2022-08-03
Unlink translations from each other

pcraig3:
# Summary | Résumé

This PR introduces a new API route and some new logic in our PageSelect component to call the new route if we select "None" as the page translation. It also adds some tests, which brings up our total coverage up to 67%.

Routes we have:

1. Set a translation: `POST` -> `http://localhost/wp-json/cds/wpml/posts/47/translation` (param: `translationId`)
2. Get a translation: `GET` -> `http://localhost/wp-json/cds/wpml/posts/47/translation`
3. (New ✨) Unset a translation: `DELETE` -> `http://localhost/wp-json/cds/wpml/posts/47/translation`

It's not technically a `DELETE`, since no resource is actually getting deleted, but Dave and I talked about it and it was the most intuitive route we came up with.

## gif

Screen Recording 2022-08-03 at 11 08 39

### DIFF
--- a/wordpress/wp-content/plugins/cds-wpml-mods/resources/js/util/fetch.js
+++ b/wordpress/wp-content/plugins/cds-wpml-mods/resources/js/util/fetch.js
@@ -1,6 +1,6 @@
 const CDS_VARS = window.CDS_VARS
 
-export const getData = async (endpoint) => {
+export const getData = async ({ endpoint }) => {
     const requestHeaders = new Headers();
     requestHeaders.append('X-WP-Nonce', CDS_VARS.rest_nonce);
 
@@ -21,14 +21,14 @@ export const getData = async (endpoint) => {
     return await response.json();
 };
 
-export const sendData = async (endpoint, data) => {
+export const sendData = async ({ endpoint, data, method }) => {
     const requestHeaders = new Headers({
         'Content-Type': 'application/json;charset=UTF-8',
     });
     requestHeaders.append('X-WP-Nonce', CDS_VARS.rest_nonce);
 
     const response = await fetch(`${CDS_VARS.rest_url}${endpoint}`, {
-        method: 'POST',
+        method: method,
         headers: requestHeaders,
         mode: 'cors',
         cache: 'default',
--- a/wordpress/wp-content/plugins/cds-wpml-mods/resources/js/util/fetch.js
+++ b/wordpress/wp-content/plugins/cds-wpml-mods/resources/js/util/fetch.js
@@ -28,7 +28,7 @@ export const sendData = async ({ endpoint, data, method }) => {
     requestHeaders.append('X-WP-Nonce', CDS_VARS.rest_nonce);
 
     const response = await fetch(`${CDS_VARS.rest_url}${endpoint}`, {
-        method: method,
+        method: method === 'DELETE' ? 'DELETE' : 'POST',
         headers: requestHeaders,
         mode: 'cors',
         cache: 'default',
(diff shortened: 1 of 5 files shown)


wordpress/wp-content/plugins/cds-wpml-mods/resources/js/util/fetch.js
```diff
@@ -21,14 +21,14 @@ export const getData = async (endpoint) => {
...
-export const sendData = async (endpoint, data) => {
+export const sendData = async ({ endpoint, data, method }) => {
     const requestHeaders = new Headers({
         'Content-Type': 'application/json;charset=UTF-8',
     });
     requestHeaders.append('X-WP-Nonce', CDS_VARS.rest_nonce);
 
     const response = await fetch(`${CDS_VARS.rest_url}${endpoint}`, {
-        method: 'POST',
+        method: method,
```
timarney, line 31: We should validate the method here so an arbitrary value isn't passed.

wordpress/wp-content/plugins/cds-wpml-mods/src/Api/Endpoints.php
```diff
@@ -191,4 +201,38 @@ public function getTranslation(WP_REST_Request $request): WP_REST_Response|WP_Er
...
+     * Get the existing translation for a given page.
+     * Returns a WP_Error if no post is found for the id given.
+     *
+     * @param  WP_REST_Request  $request
+     *
+     * @return array | WP_Error
+     */
+    public function unsetTranslation(WP_REST_Request $request)// : WP_REST_Response|WP_Error
+    {
+        $post = get_post($request['id']);
```
timarney, line 215: This should be changed to `intval($request['id'])`

timarney, line 215:
```suggestion
        $post = get_post(intval($request['id']));
```

wordpress/wp-content/plugins/cds-wpml-mods/src/Api/Endpoints.php
```diff
@@ -191,4 +201,38 @@ public function getTranslation(WP_REST_Request $request): WP_REST_Response|WP_Er
...
+
+        if (is_null($post)) {
+            return new WP_Error('post_not_found', __('The post you are looking for does not exist', 'cds-wp-mods'), array( 'status' => 404 ));
+        }
+
+        $postType = 'post_' . $post->post_type;
+        $language_code = $this->post->getLanguageCodeOfPostObject($post);
+
+        // Unset existing translation if exists
+        if ($translatedPostID = $this->post->getTranslatedPostID($post)) {
```
timarney, line 225: Might want to wrap this is a try / catch --- to ensure the `200` is actually valid

wordpress/wp-content/plugins/cds-wpml-mods/resources/js/util/fetch.js
pcraig3: Makes sense, updated.

wordpress/wp-content/plugins/cds-wpml-mods/src/Api/Endpoints.php
pcraig3: Added it everywhere we use "get_post"

wordpress/wp-content/plugins/cds-wpml-mods/src/Api/Endpoints.php
pcraig3: That works, added it to all of our calls.

timarney (approved):

PR #1014, opened 2022-08-03
Update button styles

timarney:
Looks like a "Standardized way to modify interactive states" https://github.com/WordPress/gutenberg/issues/38277 could land in a future WP release 

For now --- updating our button styles to ensure background color and border radius receive priority.

Also adds css variables to handle button background colors.

Issue: https://github.com/orgs/cds-snc/projects/14

### DIFF
--- a/wordpress/wp-content/plugins/cds-base/classes/Modules/Blocks/src/styles/editor-and-public.css
+++ b/wordpress/wp-content/plugins/cds-base/classes/Modules/Blocks/src/styles/editor-and-public.css
@@ -118,3 +118,9 @@ details[class^="wp-block"][open] {
   top: unset;
   line-height: unset;
 }
+
+/* Standardize button padding */
+.wp-block-button > .wp-block-button__link:not(.has-background),
+.wp-block-button > .wp-block-button__link.has-background {
+  padding: 12px 20px;
+}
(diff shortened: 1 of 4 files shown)


pcraig3 (approved):
Pretty good! Definitely a big improvement to how it used to work.

There's a couple of settings I noticed aren't being applied though: 
- a transparent background colour is being overruled
- the font sizes don't seem to be being applied
- default button styling is rounded in editor and square in final version (if you never set a border-radius)

| before | after |
|--------|-------|
|  some button styles in gutenberg editor      |   same button styles on a published page, with CSS rules open    |
|  Screen Shot 2022-08-03 at 13 59 44    |    Screen Shot 2022-08-03 at 13 59 29   |

pcraig3 (changes requested):
Hit "Approve" by accident, but I meant to request changes.

timarney:
Added custom sizes for font selector + css to match
Screen Shot 2022-08-03 at 3 31 32 PM

pcraig3 (approved):
Okay, I think this looks good to go, merge when you are ready 👍

timarney:
https://www.youtube.com/watch?v=_tHE1JdnqM8

PR #1033, opened 2022-08-09
Update the "create/edit new list" page and enforce filling in settings page

pcraig3:
# Summary | Résumé

This PR updates the page titles to "Mailing lists", the "create/edit new list" page, and blocks you from creating messages or lists unless you have filled out all your API settings.

| before | after |
|--------|-------|
|  Screen Shot 2022-08-09 at 09 58 22   |  Screen Shot 2022-08-12 at 10 35 31  |

Changes:
- Call them "Mailing lists" (and other content updates from Figma)
- Use the "wordpress admin" setting styles (tables where label is on the left and field is on the right)
- Remove "subscribe" and "unsubscribe" template ID fields (subscribe is now a hidden field, you enter it on the settings page)
  - "subscribe template ID" is now a hidden field pulled from the API Settings page
- Add a back link
- Important: not filling out any of the "API settings" means you can't create lists or messages
  - Error page will tell you which fields are missing

### gif of the error page

Screen Recording 2022-08-12 at 10 18 07

### DIFF
--- a/wordpress/wp-content/plugins/cds-base/classes/Modules/Notify/NotifySettings.php
+++ b/wordpress/wp-content/plugins/cds-base/classes/Modules/Notify/NotifySettings.php
@@ -116,7 +116,7 @@ class NotifySettings
 
         add_settings_section(
             'notify_api_settings_setting_section', // id
-            __('Notify API Settings', 'cds-snc'), // title
+            __('API Settings', 'cds-snc'), // title
             array( $this, 'notifyApiSettingsSectionInfo'), // callback
             'notify-api-settings-admin' // page
         );
@@ -134,7 +134,7 @@ class NotifySettings
 
         add_settings_field(
             'notify_subscribe_template', // id
-            __('Subscribe template id', 'cds-snc'), // title
+            __('Subscribe template ID', 'cds-snc'), // title
             array( $this, 'notifySubscribeTemplateIdCallback'), // callback
             'notify-api-settings-admin', // page
             'notify_api_settings_setting_section', // section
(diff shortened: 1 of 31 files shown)


wordpress/wp-content/plugins/cds-base/classes/Modules/Notify/NotifySettings.php
```diff
@@ -117,7 +116,7 @@ function ($input) {
 
         add_settings_section(
             'notify_api_settings_setting_section', // id
-            __('GC Lists Settings', 'cds-snc'), // title
+            __('Notify API Settings', 'cds-snc'), // title
```
timarney, line 119: Now that the settings are under `GC Lists` was thinking this should be GC Lists as a heading but yes they are Notify API --- what's in Figma?

wordpress/wp-content/plugins/gc-lists/resources/templates/no_api_key.php
```diff
@@ -4,16 +4,35 @@
...
 
 <h1><?php echo $title; ?></h1>
 
 <p>
     <?php
         echo sprintf(
-            __('You must configure your <a href="%s">Notify API Key</a>', 'gc-lists'),
-            admin_url("options-general.php?page=notify-settings")
+            __('You must configure your <a href="%s">Notify API Settings</a>.', 'gc-lists'),
+            admin_url("/wp-admin/admin.php?page=settings")
```
timarney, line 23: I don't think we need the `/wp-admin/` here

timarney (approved):

pcraig3:
Okay, going to merge

PR #1035, opened 2022-08-09
Add subscribe template setting

timarney:
# Summary | Résumé

Adds subscribe template setting under List Manager settings


pcraig3 (approved):
👍  looks good, works for me.

PR #1047, opened 2022-08-17
New page for choose subscribers

pcraig3:
# Summary | Résumé

This PR adds a new step after creating a new list for choosing to add subscribers or not. 
It also makes a few tweaks to our existing "Create new list" page and the "Upload subscribers" page.

## New page for subscribers

Adds the new "Add {type} subscribers" page [we see in Figma](https://www.figma.com/file/SGnOdzb51VNk2xRdA3bAJc/Bulk-sending?node-id=590%3A8450). Note that it doesn't include the images, but is based on a slightly older design (we can add images in a future PR). When it's about adding text messages, there is no subscribe block, only the option to upload.

Note that the link to "Set up a form" is meant to link to a Knowledge Base article, but it doesn't exist yet so I didn't add a link there (only applies to the "Add email subscribers").

| email subs | text message subs |
|--------|-------|
|  Screen Shot 2022-08-17 at 09 22 38    |   Screen Shot 2022-08-17 at 09 24 52  |

## Tweaks to existing pages

1.  Create new list vs. edit list
  - New lists have the "Save and continue" option, which leads to the "Add subscribers" page. Editing existing lists will just send you back to the main page of all lists.
  - New lists allow you to edit the type, but existing lists, I removed that option. I figured that we don't want to edit the type of a list after we've created it, but once we add subscribers we super don't want to change the type. So I just removed it for simplicity.
2. Upload subscribers
  - Added a back link at the top to edit the list again
  - Changed the H1
  - Added a button at the bottom to return to the mailing lists
  - Made the "choose columns/import" buttons blue since they are the primary actions

| before | after |
|--------|-------|
|        |   new back link, new h1, blue button inside the uploader    |
|   Screen Shot 2022-08-17 at 09 23 42   |  Screen Shot 2022-08-17 at 09 23 39  |

## gif

Here's the whole thing working.

Screen Recording 2022-08-17 at 09 13 29

### DIFF
--- a/wordpress/wp-content/plugins/gc-lists/resources/js/src/lists/components/ListViewTable.tsx
+++ b/wordpress/wp-content/plugins/gc-lists/resources/js/src/lists/components/ListViewTable.tsx
@@ -81,7 +81,7 @@ const EditLink = ({ listId }: { listId: string }) => {
     )
 }
 
-const UploadListLink = ({ name, listId, type }: { name: string, listId: string, type: ListType }) => {
+const UploadListLink = ({ listId, type }: { listId: string, type: ListType }) => {
     return (
         <StyledActionLink to={{ pathname: `/lists/${listId}/upload/${type}` }}>
             {__('Add subscribers', 'gc-lists')}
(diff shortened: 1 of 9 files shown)


wordpress/wp-content/plugins/gc-lists/resources/js/src/lists/components/ListViewTable.tsx
```diff
@@ -142,8 +142,8 @@ export const ListViewTable = () => {
                         <EditLink listId={`${row?.original?.id}`} />
                         <StyledDivider>|</StyledDivider>
                         {user?.isSuperAdmin && <><DeleteLink listId={`${row?.original?.id}`} /> <StyledDivider>|</StyledDivider> </>}
-                        {getListType(row?.original?.language) === ListType.EMAIL && <UploadListLink name={`${row?.values?.name}`} listId={`${row?.original?.id}`} type={ListType.EMAIL} />}
-                        {getListType(row?.original?.language) === ListType.PHONE && user?.hasPhone ? <UploadListLink name={`${row?.values?.name}`} listId={`${row?.original?.id}`} type={ListType.PHONE} /> : null}
+                        {getListType(row?.original?.language) === ListType.EMAIL && <UploadListLink listId={`${row?.original?.id}`} type={ListType.EMAIL} />}
+                        {getListType(row?.original?.language) === ListType.PHONE && user?.hasPhone ? <UploadListLink listId={`${row?.original?.id}`} type={ListType.PHONE} /> : null}
```
pcraig3, line 146: The list name doesn't get used on the upload page so I removed it.

timarney (approved):

PR #1056, opened 2022-08-18
Add parser option to rich text field

timarney:
# Summary | Résumé

Adds parser as an optional prop so that RichText fields can return structured data.

Screen Shot 2022-08-18 at 3 55 55 PM


pcraig3 (approved):
Pre-approving, but I do have a question about the uses of hyphens vs underscores. 

It looks like WP usually does hyphens but we have underscores, which might cause bugs /confusion (maybe we don't care though?). 

For example, the "related links" key is `links-related`, which is parsed out as `parsed-cds_product_links-related`, which starts with a hyphen, goes to underscores, and then ends with a hyphen.

It it's a lot of work to change it, then nvm, but since no one is using this API yet, we can still make changes without breaking anyone's workflow. Just a suggestion that all hyphens or all underscores might be cleaner.

PR #1062, opened 2022-08-22
Make menu less mysterious

pcraig3:
# Summary | Résumé

This PR attempts to clean up the screwy 3-way boolean logic that was happening with the menus.

In the site settings, there are now 3 settings for the top menu: 

1. Show Canada.ca menu
2. Show custom menu
3. No top menu

Each one does what it says, although option 2 is disabled until a custom menu has been created and is assigned to the header position. 

Resolves: https://github.com/cds-snc/gc-articles-issues/issues/481

It's probably worth getting @willeybryan's take on this before merging.

## Notes

What was happening previously was that we had a boolean option (show/hide canada.ca menu) but we actually had 3 possible states: show canada.ca menu, show custom menu, hide both menus. I wrote some logic that would try to infer intent: either show the canada.ca menu or the custom menu based on the last thing they touched, and setting "no menu"  would dequeue the custom menu behind the scenes. You can read all about it in this PR: https://github.com/cds-snc/gc-articles/pull/638

New behaviour is that anyone who never interacts with this settings page can toggle between the default Canada.ca menu and a custom menu as before, but you can also be explicit by selecting one of the 3 options in Site Settings. Should make things clearer, and nothing gets dequeued unexpectedly now.

## Screenshots

I also added `<fieldset>`s to the lists of radios and now they have way more space. 

| before | after (no custom menu assigned) | after (custom menu assigned) |
|--------|---------------------------------|------------------------------|
|   Current settings page. Radios are closer together and there is only 2 options for top menu.     |        More space between radios. "Set custom menu" is disabled and there is a link to the menu page.         |     More space between radios. "Set custom menu" is enabled and the name of the custom menu is displayed       |
|   Screen Shot 2022-08-22 at 18 42 43    |             Screen Shot 2022-08-22 at 18 42 47               |  Screen Shot 2022-08-22 at 18 44 27     |

### DIFF
--- a/wordpress/wp-content/plugins/cds-base/classes/Modules/Blocks/src/styles/public.css
+++ b/wordpress/wp-content/plugins/cds-base/classes/Modules/Blocks/src/styles/public.css
@@ -102,8 +102,12 @@ header #wb-srch .submit {
   display: none;
 }
 
-/* Hide menu (keeps the blue bar) */
-.hide-wet-menu .gcweb-menu .container {
+/* Hide Canada.ca menu (including the blue bar) */
+.hide-wet-menu .gcweb-menu {
+  display: none;
+}
+
+.hide-custom-menu .nav--primary__container {
   display: none;
 }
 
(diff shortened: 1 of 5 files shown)


willeybryan:
Oh that's cool, we show the name of their menu for more context, I like that.

timarney (approved):

PR #1071, opened 2022-08-24
Patch for the missing HTTP_REFERER issue

pcraig3:
# Summary | Résumé

Adding a patch that I think will resolve this checklists bug we're seeing in staging and prod: https://github.com/cds-snc/gc-articles-issues/issues/490

The idea is that if we are on the publishpress page that we prefill the `$_SERVER['HTTP_REFERER']` key with the current page URL. It's hard to test locally but I think it will do the trick.

also opened up an issue here: https://github.com/publishpress/PublishPress-Checklists/issues/411


timarney (approved):
Will need to test on staging as we can't re-create locally

PR #1089, opened 2022-09-21
fix: add encrypted GitHub token for Composer

patheard:
# Summary
Update the Renovate config to include an encrypted GitHub token that will allow Renovate to clone Composer dependency GitHub repositories without any rate limiting.

Token was encrypted and scoped to only work on cds-snc/gc-articles using the Renovate bot's public key:
https://docs.renovatebot.com/configuration-options/#encrypted

# Related
* #1024 
* #1025 
* cds-snc/renovate-config-private#2
* renovatebot/renovate#16193

PR #1091, opened 2022-09-21
fix: update config extends to use `local`

patheard:
# Summary
Update the Renovate `extends` to use `local` to include config.  This should then allow Renovate to load the configuration from internal org repos.

With the current setup, Renovate is failing to load the private config as its token does not have access:
```
DEBUG: Failed to retrieve config/gc-articles.json from repo
{
  "statusCode": 404,
  "url": "https://api.github.com/repos/cds-snc/renovate-config-private/contents/config/gc-articles.json"
}
```

# Related
* #1089 
* [Renovate private config presets](https://docs.renovatebot.com/getting-started/private-packages/#private-config-presets)

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -1,8 +1,8 @@
 {
   "$schema": "https://docs.renovatebot.com/renovate-schema.json",
   "extends": [
-    "github>cds-snc/renovate-config",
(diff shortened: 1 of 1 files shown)


dinophile (approved):
🍌

ascheid (approved):

PR #1102, opened 2022-10-31
refactor: use user returned from wp_set_current_user

jeberhardt:
# Summary | Résumé

There is no issue for this PR.

A small refactor, using the `user` object returned from `wp_set_current_user`, rather than setting the user, and then retrieving it via `wp_get_current_user`.

https://developer.wordpress.org/reference/functions/wp_set_current_user/

Since all the code changed in this PR is found in tests, then to test this PR, just run all the tests found in `~/wordpress/wp-content/plugins/gc-lists`


dsamojlenko (approved):
LGTM

PR #1113, opened 2022-11-02
test: add verbose flag to GC Lists Pest tests

patheard:
# Summary
Update the PHP Pest test command for GC Lists to provide verbose output.


patheard:
No extra info provided by the flag.

PR #1127, opened 2022-11-15
chore: add tag for security-tools

jeberhardt:
# Summary | Résumé

This PR is to pin the `cds-snc/security-tools` package to the tag `v1`. This will prevent Renovate from trying to update it every time there is any kind of update to the `security-tools` repo.


patheard (approved):
For science 🧪

PR #1129, opened 2022-11-16
fix: category language switcher

jeberhardt:
# Summary | Résumé

Closes: cds-snc/gc-articles-issues#525

This PR is a hack, definitely. There is a bug in WPML that incorrectly returns the URL that is supposed to be for the translated page. It returns the `translated_url` for the same language, instead of the url for the translated content.

- go to a "category" page (_uncategorized_, for example.
- click the language switcher link at the top-right side of the page
- it should switch languages for that category page
- the language switcher should continue to work as before on all other pages

### DIFF
--- a/wordpress/wp-content/themes/cds-default/inc/template-functions.php
+++ b/wordpress/wp-content/themes/cds-default/inc/template-functions.php
@@ -227,7 +227,7 @@ function language_switcher_output($languages)
         foreach ($languages as $language) {
             $text = get_language_text($language['translated_name']);
             if (!$language['active']) {
-                $link = '<a lang="' . $text['abbr'] . '" hreflang="' . $text['abbr'] . '" href="' . check_translated_url($language['url'], $text['abbr']) . '">';
+                $link = '<a lang="' . $text['abbr'] . '" hreflang="' . $text['abbr'] . '" href="' . convert_url($language['url'], $text['abbr']) . '">';
                 $link .= '<span class="hidden-xs">' . $text['full'] . '</span>';
                 $link .= '<abbr title="' . $text['full'] . '" class="visible-xs h3 mrgn-tp-sm mrgn-bttm-0 text-uppercase">';
                 $link .= $text['abbr'];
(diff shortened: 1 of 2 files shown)


wordpress/wp-content/themes/cds-default/inc/template-functions.php
```diff
@@ -244,6 +244,55 @@ function language_switcher_output($languages)
...
+    *
+    *                   Wordpress will then redirect to the correct url for the target language:
+    *                       ~/fr/category/test-fr
+*/
+function convert_url($url, $lang): string
+{
+    $category_path_name = "category";
+    $return_url = $url;
+    $parsed_url = parse_url($return_url);
+    if (str_contains($parsed_url['path'], "/$category_path_name/")) { // only modify the url if it is a category page
```
timarney, line 287: Haven't tested this PR yet but might suggest getting rid of the else

```php
if (str_contains($parsed_url['path'], "/$category_path_name/")) { // only modify the url if it is a category page
    if (str_starts_with($parsed_url['path'], "/$category_path_name/")) { // currently on the EN path
        $return_url = str_replace("/$category_path_name/", "/$lang/$category_path_name/", $return_url);
    } 
    
    // otherwise it must be on the FR path
    return $return_url = str_replace("/fr/", "/", $return_url);
        
}
```

wordpress/wp-content/themes/cds-default/inc/template-functions.php
```diff
@@ -244,6 +244,56 @@ function language_switcher_output($languages)
 
     return $langs;
 }
+/*
+    TODO: Replace this hack function with a proper solution.
```
timarney, line 248: Let's call this a workaround vs ` total hack` --- 

 total hack doesn't inspire confidence in the product if folks happen to be looking at the source code

jeberhardt: Fair enough. I wanted to impress the importance of finding a better long term solution.

timarney (approved):
Approved - can get this pushed up to staging for further team review.

PR #1158, opened 2023-01-12
fix: add filter for core-image

jeberhardt:
# Summary | Résumé

Fixes cds-snc/gc-articles-issues#552

> Create an article
> Add an image using the block editor
> Set the size of the image to "Full"
> Load the page and resize the browser window
> Observe that the image maintains the correct aspect ratio


timarney (approved):
Didn't test locally but code looks good

PR #1182, opened 2023-03-03
fix: set WAF ACL body oversize handling

patheard:
# Summary
Update the WAF ACL rules with `body` blocks to set the oversize request body size handling to `CONTINUE` with inspection.

This is required by AWS to continue to allow API management of WAF ACLs.

Also updates the Terraform provider to the latest version to provide support for the new `oversize_handling` attribute.

# Related
- cds-snc/gc-articles-issues#531


CalvinRodo (approved):
:taco::cat:

dsamojlenko (approved):

PR #1184, opened 2023-04-24
fix: remove empty EFS security group

patheard:
# Summary
Update the Terraform EFS security group logic so that it is only created if EFS is enabled.


maxneuvians (approved):

CalvinRodo (commented):
🤦🏼 :taco::cat:

PR #1190, opened 2023-05-02
fix: set EFS security group ID to blank

patheard:
# Summary 
Update how the EFS security group ID is set from the `network` module.  This is because a `null` value removes the output and causes an error with the `efs` module.


CalvinRodo (approved):
:taco::cat:

maxneuvians (approved):

PR #1192, opened 2023-05-02
fix: disable WPML auto updates

patheard:
# Summary 
Update the WP config to prevent WPML plugin from
attempting to auto-update.  This will also remove
the error appearing in the admin panel.

image

# Related
- cds-snc/platform-core-services#311


CalvinRodo (approved):
:taco::cat:

PR #1199, opened 2023-05-05
chore: update WPML Wordpress plugins

patheard:
# Summary
Update the WPML plugins to their latest versions.

Pin to Node v16 in the devcontainer.

# Related
- cds-snc/platform-core-services#311


CalvinRodo (approved):
:taco::cat:

PR #1207, opened 2023-05-08
Version Bump 3.9.5

patheard:
# Summary
Fix site creation and update CSP to fix the NYN content violation.

# Related
- #1206


CalvinRodo (approved):
:taco::cat:

PR #1211, opened 2023-05-09
Version Bump 3.9.6

patheard:
# Summary
Fixes `.is-flex-layout` CSS and adds CSP SHAs for the Yoast plugin.


CalvinRodo (approved):
:taco::cat:

PR #1222, opened 2023-05-09
fix: Docker SBOM generate step

patheard:
# Summary
Update the Docker SBOM step to run before the ECR
logout, as an active ECR session is now required.


CalvinRodo (approved):
:taco::cat:

PR #1240, opened 2023-05-19
feat: replace Canary with Route53 health check

patheard:
# Summary
Add a Route53 health check to replace the CloudWatch Canary.

Also update the WAF custom body size rule with the new 16kb limit for CloudFront WAF ACLs.


CalvinRodo (approved):
:taco::cat:

I like the diffs on this one.

PR #1285, opened 2023-06-26
fix: invalidate entire sub-site cache

patheard:
# Summary
Add a filter to the CDS default them that alters the behaviour of the `c3-cloudfront-clear-cache` plugin so that cache invalidation requests cause the entire subsite's CloudFront cache to be cleared.

This is to work around an existing bug in the plugin that is preventing any cache invalidations from running.  There are several open issues and PRs on the plugin to fix the issue but they haven't been merged or triaged.

# Related
- https://github.com/cds-snc/platform-core-services/issues/401


CalvinRodo (approved):
amazing

PR #1286, opened 2023-06-27
fix: add null checking and prevent direct access

patheard:
# Summary
Update the CDS Default theme to add `null` checking and prevent direct access to the theme's `index.php` file.

More thorough direct PHP access blocks will be added in a subsequent PR.

# Related
- https://github.com/cds-snc/platform-core-services/issues/403
- https://github.com/cds-snc/platform-core-services/issues/404


CalvinRodo (approved):
:taco::cat:

PR #1292, opened 2023-06-29
fix: WordPress error metric filter pattern

patheard:
# Summary
Update the error metric filter pattern to use a syntax that allows for multiple optional tokens while also ignoring a specific token.

# Related
- https://github.com/cds-snc/platform-core-services/issues/397


CalvinRodo (approved):
:taco::cat:

PR #1296, opened 2023-06-29
fix: add Sentinel subscription filter pattern

patheard:
# Summary
Add the CloudWatch log group subscription filter that will cause all WordPress ECS logs to be forwarded
to Sentinel.

# Related
- https://github.com/cds-snc/platform-core-services/issues/398


CalvinRodo (approved):
:taco::cat:

PR #1320, opened 2023-07-12
fix: locust load tests and devcontainer

patheard:
# Summary
Update the locust load tests and devcontainer so
that they run properly.

# Related
- https://github.com/cds-snc/platform-core-services/issues/415


CalvinRodo (approved):
:taco::cat:

PR #1324, opened 2023-07-13
chore: more error tokens to ignore

patheard:
# Summary
Update the WordPress CloudWatch `error` metric filters with more tokens that should be ignored.  These correspond to:

- People attempting to generate directory listings.
- People resetting their password.

# Related


CalvinRodo (approved):
:taco::cat:

PR #1328, opened 2023-07-16
chore: suppress CloudWatch error false-positives

patheard:
# Summary
Update the CloudWatch error metric filter to ignore `GET` requests to the Notify wp-json endpoint.

This is because these will often contain error tokens as part of the `GET` request parametres during fuzzing attacks.

# Related


CalvinRodo (approved):
:taco::cat:

PR #1335, opened 2023-07-17
fix: add password reset false-positive suppress

patheard:
# Summary
Update the CloudWatch `error` metric filter so that it does not trigger for password reset requests.

# Related
- https://github.com/cds-snc/platform-core-services/issues/425


CalvinRodo (approved):
:taco::cat:

PR #1337, opened 2023-07-19
Allow json files to be uploaded

dsamojlenko:
# Summary | Résumé

Allow JSON type files to be uploaded


timarney (approved):
Approving this - 

We'll be using it on the forms team

PR #1346, opened 2023-07-21
fix: ignore WordPress password GET requests

patheard:
# Summary
Update the CloudWatch `error` metric filter to ignore all password reset requests that include the word `error`.


CalvinRodo (approved):
:taco::cat:

PR #1348, opened 2023-07-21
fix: use SRE bot slack notifications

patheard:
# Summary
Update do use the SRE bot slack notification rather than the TF module.


CalvinRodo (approved):
:taco::bot:

PR #1353, opened 2023-07-24
chore: ignore CloudWatch errors from 404 requests

patheard:
# Summary
Update the CloudWatch `error` metric filter so that it doesn't count HTTP requests that have a status of 404.

This is being done to stop fuzzing attacks from triggering an error alarm.


gcharest (approved):

wanpengyang (commented):
:shipit:

PR #1363, opened 2023-07-31
fix: protect against null `$post` object

patheard:
# Summary
Update the excerpt function to check if the `$post` object is null before attempting to create a "read more" link.

# Related
- https://github.com/cds-snc/platform-core-services/issues/433


sylviamclaughlin (approved):

wanpengyang (approved):
👍

PR #1379, opened 2023-08-28
fix: gc-lists plugin dev dependencies

patheard:
# Summary
Add the `@babel/plugin-proposal-private-property-in-object` dependency which was being implicitly provided by another dependency.


CalvinRodo (approved):
:taco::cat:

PR #1388, opened 2023-08-30
feat: log non-Government of Canada logins

patheard:
# Summary
Update the `cds-base` plugin so that when a user with a non-GC email address successfully authenticates, a log is written.

This will allow us to alarm on unexpected email address logins.

# Related
- https://github.com/cds-snc/platform-core-services/issues/439


CalvinRodo (approved):
:taco::cat:

PR #1422, opened 2023-09-18
chore: use OIDC for AWS auth from GitHub actions

patheard:
# Summary
Update the GitHub actions that need to authenticate with an AWS account to use OIDC roles.

# Related
- https://github.com/cds-snc/platform-core-services/issues/459


CalvinRodo (approved):
:taco::cat:

PR #1427, opened 2023-09-21
chore: WordPress database error CloudWatch alarm

patheard:
# Summary
Add a CloudWatch alarm for WordPress database alarms and ignore the term `database error` in the default WordPress error alarm.

This is being done because there are some unpatched plugins in use that generate database errors with no user facing impact.  This new alarm will ignore these but still catch cases where there is a serious database issue.

# Related
- https://github.com/cds-snc/platform-core-services/issues/465


CalvinRodo (approved):
:taco::cat:

PR #1446, opened 2023-09-26
fix: allow OIDC ECR role to pull images

patheard:
# Summary
Update the OIDC ECR role's permission policy to
allow it to pull images as well.

This is needed for the Docker SBOM workflow step.

# Related
- https://github.com/cds-snc/platform-core-services/issues/469


CalvinRodo (approved):
:taco::cat:

PR #1448, opened 2023-09-26
chore: fix deprecated workflow set-output commands

patheard:
# Summary
Update the deprecated `set-output` GitHub workflow commands to use the [new syntax](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-output-parameter).


CalvinRodo (approved):
:taco::cat:

PR #1452, opened 2023-09-26
fix: top bar menu item small screen margin

patheard:
# Summary
Fix the top bar menu item CSS so that the margin does not push the link out of alignment along the right edge of the screen.

# Related
- https://github.com/cds-snc/platform-core-services/issues/468


CalvinRodo (approved):
:taco::cat:

PR #1471, opened 2023-10-06
fix: Google Tag Manager inline CSP hash

patheard:
# Summary
Fix the hash used by the Content Security Policy to allow the Google Tag Manager inline script to run.

# Related
- https://github.com/cds-snc/platform-core-services/issues/477


wanpengyang (approved):
👍

PR #1538, opened 2023-11-24
fix: add CSP hashes for Google Analytics

patheard:
# Summary
Add the CSP hashes for the GC Data Conference sub-site.


CalvinRodo (approved):
:taco::cat:

PR #1570, opened 2023-12-22
chore: ignore invalid email address error from ListManager

patheard:
# Summary
Update the CloudWatch error metric so that when an invalid email address error is returned from ListManager, it does not trigger an alarm.


CalvinRodo (approved):
:taco::cat:

PR #1580, opened 2024-01-18
chore: suppress HTTP 301 CloudWatch alarms

patheard:
# Summary
Update the CloudWatch error alarms so that they do no trigger for HTTP 301 request logs.

This is being done because of XSS scanners commonly using the `onerror` tag, which triggers the alarm.


wanpengyang (approved):
👍

PR #1737, opened 2024-05-08
fix: update RDS module to use standalone SG rules

patheard:
# Summary
Update to the latest RDS module which includes a change to standalone security groups.

This will stop the custom VPN security group rule from being recreated on each apply.

# Related
- https://github.com/cds-snc/terraform-modules/pull/481


CalvinRodo (approved):
:taco::cat:

PR #2087, opened 2025-02-24
chore: add redirect for National Manager's Community

patheard:
# Summary
Add a redirect to send NMC requests to their new site.  This redirect includes logic to still allow access to the WordPress admin dashboard so that GC Lists can still be used in the short term.

# Related
- https://github.com/cds-snc/platform-core-services/issues/676


willeybryan (approved):
LGTM

PR #2148, opened 2025-04-29
chore: remove unused plugins

patheard:
# Summary
Remove plugins that are no longer required.  This is in preparation for converting Articles to a headless CMS.

# Related


willeybryan (commented):
I also see some packages being removed like:
https://en-ca.wordpress.org/plugins/wp-native-php-sessions/
wpackagist-plugin/gutenberg-custom-fields
wpackagist-plugin/simple-history


Have we double checked that nothing we are keeping uses those packages?

patheard:
This is what I'd like to do on Staging.  I've confirmed that locally everything works without the features I've removed, but I'd like to validate it with more of our site content before going to Prod.

willeybryan (approved):
Oh, yeah, smart. I totally remembered that's a thing.  😅
LGTM

PR #2154, opened 2025-05-02
fix: display of breadcrumbs for nested pages

patheard:
# Summary
Update the breadcrumb function so that it correctly shows parent page links and does not provide an entry for the current page.

This rework is required because previously the breadcrumbs were being generated by the Yoast SEO plugin.

# Related


CalvinRodo (approved):
:taco::cat:

PR #2161, opened 2025-05-07
chore: remove GC Lists plugin

patheard:
# Summary
Remove the GC Lists plugin from Articles.

# Related


CalvinRodo (approved):
:taco::cat:

PR #2195, opened 2025-06-06
chore: remove PrivateLink VPC endpoints

patheard:
# Summary
Update to remove the PrivateLink VPC endpoints as these are no longer required and are expensive to maintain.

The purpose of the PrivateLink endpoints was to keep AWS traffic to those services within the AWS network.  However, they cost ~ 300 USD / month for both environments and are not required to meet security requirements as all traffic flowing to those services is unclassified and encrypted-in-transit.

# Related


willeybryan (approved):
lgtm

PR #2270, opened 2025-08-25
fix: default theme WET footer markup

patheard:
# Summary
Fix the display of the WET footer in the CDS default theme.  This was caused by an upstream change to the CSS and associated footer class names.

# Related
- https://github.com/cds-snc/platform-core-services/issues/814


willeybryan (approved):
LGTM

PR #2557, opened 2026-04-13
feat: setup DNS proxy safelist

patheard:
# Summary
Update the DNS proxy action to the latest version and set it to use a domain safe-list.

This will block queries to unexpected domains from the workflows.

# Related
- https://github.com/cds-snc/platform-core-services/issues/972


willeybryan (approved):
lgtm

PR #2560, opened 2026-04-14
feat: added webhook file and updated functions

omartehsin1:
# Summary | Résumé

Related to: https://github.com/cds-snc/platform-core-services/issues/962

Created webhook functionality in the default theme. Integrated with theme, so webhook now triggers on post publish for the default theme.

### DIFF
--- a/wordpress/wp-content/themes/cds-default/inc/webhook.php
+++ b/wordpress/wp-content/themes/cds-default/inc/webhook.php
@@ -1,0 +1,38 @@
+<?php
+
+function onSavePost($post_ID, $post)
+{
+    if ($post->post_status === "publish") {
+        try {
+            $url = "https://api.github.com/repos/cds-snc/cds-website-pr-bot/dispatches";
+            $token = get_option('GITHUB_AUTH_TOKEN');
+
+            if (empty($token)) {
+                return;
+            }
+
+            $args = [
+                'headers' => [
+                'Accept' => 'application/vnd.github+json',
+                'Authorization' => 'token ' . $token,
+                'Content-Type' => 'application/json'
+                ],
+                'body'        => json_encode(['event_type' => 'strapi_update'])
+            ];
+
+            $response = wp_remote_post($url, $args);
+
+            if (is_wp_error($response)) {
+                $error_message = $response->get_error_message();
+                error_log("Github :: Something went wrong: $error_message");
+            } else {
+                error_log('Github :: response:');
+                error_log(print_r($response['response'], true));
+            }
+        } catch (Exception $e) {
+            error_log('Caught exception: ' . $e->getMessage() . "\n");
+        }
+    }
+}
+
+add_action('save_post', 'onSavePost', 10, 2);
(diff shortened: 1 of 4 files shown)


wordpress/wp-content/themes/cds-default/inc/webhook.php
```diff
@@ -0,0 +1,38 @@
+<?php
+
+function onSavePost($post_ID, $post)
+{
+    if ($post->post_status === "publish") {
+        try {
+            $url = "https://api.github.com/repos/cds-snc/cds-website-pr-bot/dispatches";
```
patheard: Can you make this an input parameter as well so that teams can set the URL of the repo that want to trigger a workflow in.

wordpress/wp-content/themes/cds-default/inc/webhook.php
```diff
@@ -0,0 +1,46 @@
+<?php
+
+function onSavePost($post_ID, $post)
+{
+    if ($post->post_status === "publish") {
+        try {
+            // Get configurable repository URL
+            $repo_url = get_option('GITHUB_REPOSITORY_URL', 'cds-snc/cds-website-pr-bot');
```
patheard: Nice! Could you remove the default value behaviour though since I'd like the user to explicitly provide both values.

wordpress/wp-content/themes/cds-default/settings.php
```diff
@@ -0,0 +1,173 @@
...
+        }
+    );
+
+    // Register GitHub Repository URL setting
+    register_setting(
+        'cds_default_github_option_group',
+        'GITHUB_REPOSITORY_URL',
+        function ($input) {
+            if ($input == '') {
+                return get_option('GITHUB_REPOSITORY_URL', 'cds-snc/cds-website-pr-bot');
```
patheard: Same comment here about removing the default value
```suggestion
                return get_option('GITHUB_REPOSITORY_URL');
```

patheard (changes requested):

wordpress/wp-content/themes/cds-default/settings.php
```diff
@@ -0,0 +1,173 @@
...
+
+    printf(
+        '<input class="regular-text" type="text" name="GITHUB_AUTH_TOKEN" id="github_auth_token" aria-describedby="github_auth_token_value" value="" placeholder="ghp_xxxxxxxxxxxxxxxxxxxx">'
+    );
+    echo '<p class="description">' . __('GitHub personal access token with repo permissions.', 'cds-snc') . '</p>';
+}
+
+function cds_default_repository_callback($args)
+{
+    $current_repo = get_option('GITHUB_REPOSITORY_URL', 'cds-snc/cds-website-pr-bot');
```
patheard: Just tested and realized this one needs to change as well:
```suggestion
    $current_repo = get_option('GITHUB_REPOSITORY_URL');
```

wordpress/wp-content/themes/cds-default/settings.php
```diff
@@ -0,0 +1,173 @@
...
+    );
+
+    // Register GitHub Repository URL setting
+    register_setting(
+        'cds_default_github_option_group',
+        'GITHUB_REPOSITORY_URL',
+        function ($input) {
+            if ($input == '') {
+                return get_option('GITHUB_REPOSITORY_URL');
+            }
```
patheard: One last change 🙏🏻 
```suggestion
```
This will allow the user to unset the repo URL if they want to turn off the behaviour

patheard (approved):

PR #2581, opened 2026-04-18
fix: WAF improvements based on review

patheard:
# Summary
Adjust the WAF rules based on a review with AWS. This includes the following:

1. Optimize rule order. Cheaper rules should come first.
2. Add InvalidHost and Anti-DDoS rules.
3. Set BotControl to block with rule overrides to allow headless CMS use.
4. Remove the ALB WAF and use security groups to only allow ingress to the ALB from CloudFront.

# Related
- https://github.com/cds-snc/platform-core-services/issues/955


willeybryan (approved):
LGTM

PR #2597, opened 2026-04-27
fix: pass Docker secrets to WordPress image

patheard:
# Summary
Update the SBOM and vulnerability scan so that the WPML secrets are passed in to allow the dependency download.


willeybryan (approved):
LGTM

PR #2606, opened 2026-05-05
Feat: update firewall rules

omartehsin1:
# Summary | Résumé

- Reordered rules by WCU lowest to highest


infrastructure/terragrunt/aws/load-balancer/waf.tf
```diff
@@ -1057,7 +1057,7 @@ resource "aws_wafv2_web_acl" "wordpress_waf" {
 
   rule {
     name     = "BotControl"
-    priority = 210
+    priority = 70
```
patheard: Could you reorder this rule to be last to be executed. It has an extra cost of $1/million requests scanned so ideally it runs the least out of all of them

patheard (approved):

PR #2610, opened 2026-05-06
fix: increase the WAF GET rate limits

patheard:
# Summary
Update the IP and JA4 rate limits allow 2,000 requests in a 5 minute period. We'll revert this once we get the Notify admin header secret added as a bypass.

Note this has already been updated in Prod through the console.


willeybryan (approved):
LGTM

PR #2618, opened 2026-05-14
feat: added docker push deploy oidc roles

omartehsin1:
# Summary | Résumé

- Scoped down roles
- Created `gc-articles-docker-push` and `gc-articles-docker-deploy`
- Updated workflows

Related to: https://github.com/orgs/cds-snc/projects/51?pane=issue&itemId=164542666&issue=cds-snc%7Cplatform-core-services%7C958

### DIFF
--- a/infrastructure/terragrunt/aws/ecr/oidc.tf
+++ b/infrastructure/terragrunt/aws/ecr/oidc.tf
@@ -1,5 +1,6 @@
 locals {
   ecr_tag_release = "gc-articles-ecr-tag-release"
+  docker_push     = "gc-articles-docker-push"
 }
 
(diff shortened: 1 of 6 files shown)


infrastructure/terragrunt/aws/ecr/oidc.tf
patheard: Instead of adding a new `module "docker_push"` below, you can just add the new role as another array item like so:
```suggestion
    },
    {
      name      = local.docker_push
      repo_name = "gc-articles"
      claim     = "ref:refs/heads/main"
    }
```

patheard (approved):

PR #2628, opened 2026-05-20
fix: update JA4 rules to challenge

patheard:
# Summary
Update the JA4 rate based rules to `challenge` rather than `block` when there is a match. This will allow legitimate browser traffic through in the event of fuzzing attacks sharing the same JA4 fingerprint.

Additionally update the JA4 to have `NO_MATCH` behaviour if a fingerprint cannot be determine to prevent accidental aggregation with legitimate traffic.

Finally, fix a bug with the BotControl scope down statement.

# Related
- https://github.com/cds-snc/platform-core-services/issues/1003


willeybryan (approved):
LGTM

PR #2691, opened 2026-06-25
fix: remove leading space from class attribute

patheard:
# Summary
Update the `wp-block-paragraph` filter so that leading spaces are also removed from the `class` attribute when the className is deleted. This fixes the cases where multiple CSS classes were on the `<p>` tag.


willeybryan (approved):
LGTM

